PR #1, opened 2015-05-07
Rewritten ptototype.query method

nucleusv:
This method now uses the same request doZabbixAPIRequest without code
dublication. Please check on your installation cause I have only one
metric per graph

I thinl idea is clear to remove code dublication with auth, and do all requests thru doZabbixAPIRequest

### DIFF
--- a/zabbix/datasource.js
+++ b/zabbix/datasource.js
@@ -56,20 +56,11 @@ function (angular, _, kbn) {
       from = Math.ceil(from/1000);
       to = Math.ceil(to/1000);
 
-      var performedQuery;
+           
 
-      // Check authorization first
-      if (!this.auth) {
-        var self = this;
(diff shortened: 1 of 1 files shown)


nucleusv:
Alex, is it possible in grafana by default replace word alias with the metric name from zabbix api, and add check box "autoupdate metric name fro api" if it is set to true, than every time when graph is loaded to take metric name from api, not from alias field? It more convenient for zabbix users.

alexanderzobnin:
It's good idea. I'll try to realize this feature later.

nucleusv:
Have you fully tested this patch ?
Is it work for you from every beginning? Cause I see in console there errors sometimes.
There is cosole.log().

alexanderzobnin:
Yes, I check for errors and for my installation all right. if you see errors again, please open issue with detailed error description.

PR #2, opened 2015-05-08
Added incrimented field for each request

nucleusv:
Added incrementing requestId variable for each request, not always write good due to async requests.

Removed condole.log()

### DIFF
--- a/zabbix/datasource.js
+++ b/zabbix/datasource.js
@@ -28,6 +28,8 @@ function (angular, _, kbn) {
 
       this.supportMetrics   = true;
       this.supportAnnotations = true;
+	  
+	  this.requestId = 0;
     }
 
 
@@ -56,12 +58,9 @@ function (angular, _, kbn) {
       from = Math.ceil(from/1000);
       to = Math.ceil(to/1000);
 
-           
-
       return this.performTimeSeriesQuery(target_items, from, to).then(function (response) {
 		  
-		  console.log(response);
-        /**
+		 /**
          * Response should be in the format:
          * data: [
          *          {
@@ -182,7 +181,7 @@ function (angular, _, kbn) {
               time_from: start,
           },
           auth: this.auth,
-          id: 1
+          id: ++this.requestId
     };
       
       // Relative queries (e.g. last hour) don't include an end time
@@ -207,7 +206,7 @@ function (angular, _, kbn) {
             password: this.password
           },
           auth: null,
-          id: 1
+          id: ++this.requestId
         },
       };
 
@@ -230,7 +229,7 @@ function (angular, _, kbn) {
           sortfield: 'name'
         },
         auth: this.auth,
-        id: 1
+        id: ++this.requestId
       };
 
       return this.doZabbixAPIRequest(data);
@@ -247,7 +246,7 @@ function (angular, _, kbn) {
           sortfield: 'name'
         },
         auth: this.auth,
-        id: 1
+        id: ++this.requestId
       };
       if (groupid) {
         data.params.groupids = groupid;
@@ -268,7 +267,7 @@ function (angular, _, kbn) {
           hostids: hostid
         },
         auth: this.auth,
-        id: 1
+        id: ++this.requestId
       };
 
       return this.doZabbixAPIRequest(data);
@@ -286,7 +285,7 @@ function (angular, _, kbn) {
           hostids: hostid
         },
         auth: this.auth,
-        id: 1
+        id: ++this.requestId
       };
       // If application selected return only relative items
       if (applicationid) {
@@ -316,7 +315,7 @@ function (angular, _, kbn) {
               limit: self.limitmetrics,
           },
           auth: self.auth,
-          id: 1
+          id: ++self.requestId
         },
       };
 
@@ -339,7 +338,7 @@ function (angular, _, kbn) {
                 limit: self.limitmetrics,
             },
             auth: self.auth,
-            id: 1
+            id: ++self.requestId
           },
         };
 


alexanderzobnin:
I have already removed console.log(). Please, fetch your repo for new changes.

alexanderzobnin:
Note about "id" parameter from [Zabbix API docs](https://www.zabbix.com/documentation/1.8/api/getting_started)
> “id”: 1 - this field can be used to tie every JSON request to it's response. The response will have the same “id” as provided by the request. It is useful when you are sending multiple requests at once. These are not required to be unique or sequential;
For this reason I don't use increment or other id calculation.

nucleusv:
Alex, look at errors,
try yo create a new graph and add metrics.

I add new metrics, but I have only one line 
https://yadi.sk/i/r_VaNA4agWkL5
https://yadi.sk/i/oGGMCsDBgWkQD
https://yadi.sk/i/6_DTlPligWkRJ

alexanderzobnin:
I think this is because different data types in one graph. I note this moment in code:
```
 ZabbixAPIDatasource.prototype.performTimeSeriesQuery = function(items, start, end) {
      var item_ids = items.map(function (item, index, array) {
        return item.itemid;
      });

      // TODO: if different value types passed?
      //       Perform multiple api request.
      var history_type = items[0].value_type;
```
I'll try to fix this problem in this release.

alexanderzobnin:
Need to group requests with same data type and do multiple api request. Just with incremental id ))

nucleusv:
I think to use incremental Id is not best way, that is way id field must be unique, I think it is better to use Math.random() * 10000 function ?

alexanderzobnin:
> “id”: 1 - this field can be used to tie every JSON request to it's response. 

id must be unique within one http request. We may use for example id: 1 for integer data and id: 2 for float.

nucleusv:
To be honest, I think it is not needed in angularjs apps, you need to rewrite function history get that do async request to api, and when all subrequest functions return result merge them in one array and return.

nucleusv:
In function 
     ZabbixAPIDatasource.prototype.performTimeSeriesQuery = function(items, start, end)

you need to run over all array items, and create new hash  history_type[type] = array_of_items 

then you need to run over history_type hash and

for in  history_type as key=> value {

do data json for each request 
performQuery = function(){ this.doZabbixAPIRequest(data1) } for the first hash element 

and then 

performQuery.then(function(){this.doZabbixAPIRequest(dataN) })

}

return perfromQuery()

////
It is just a logic not a code :)

alexanderzobnin:
Ok, I thought the same. I'll do it tomorrow.

nucleusv:
Great!

And for future I hope this patch will be in zabbix release soon: https://support.zabbix.com/browse/ZBXNEXT-1193

If we ask data more than 24 hours we should take data from trend(s).get method instead from history.get
What do you think?

alexanderzobnin:
Yes, I also notice this problem. It complicated work with long-time graphs. This patch will solve problem.

alexanderzobnin:
I fixed bug with different data types in one graph. See release-1.0 branch.

nucleusv:
Alex, your code like a charm :+1:  I will test it tommorrow.

Another thought have come to my mind:

According to data types in zabbix
     `<select class="input select" id="value_type" name="value_type" size="1">`
        `<option value="3" selected="selected">Числовой (целое положительное)</option>`
       ` <option value="0">Числовой (с плавающей точкой)</option>`
       `<option value="1">Символ</option><option value="2">Журнал (лог)</option>`
       `<option value="4">Текст</option>`
   `</select>`

This method must return only items wich has a value type of float or integer:

    ZabbixAPIDatasource.prototype.performItemSuggestQuery 
        
    params: {
          output: ['name', 'key_', 'value_type', 'delay'],
          sortfield: 'name',
          hostids: hostid,
          filter: {
                     value_type: [0,3]
          }
        }
      
from zabbix docs:

filter	- object	
Return only those results that exactly match the given filter.

Accepts an array, where the keys are property names, and the values are either a single value or an array of values to match against. 

Supports additional filters: 
host - technical name of the host that the item belongs to.

https://www.zabbix.com/documentation/2.4/manual/api/reference/item/get
      

`

alexanderzobnin:
Thanks. It is a good idea for graphs - non-numeric items just not needed.

nucleusv:
Alex, do you have your own zabbix installation? 
Can you add trends api patch to it?
If not I will create VM on ditigatal patch for testing ?

nucleusv:
Any success with version for grafana-2 ?

alexanderzobnin:
I have production zabbix server but in isolated network without Internet access.
In Grafana-2 I have some troubles with configuration page and data collecting via grafana backend. Plugin work only in grafna-1.9 like style (request data via frontend). In other respects this version very similar inside.

nucleusv:
Alex, please add filter to return only host groups with real servers, host groups with templates only are not needed I thinl

    // Get the list of host groups
    ZabbixAPIDatasource.prototype.performHostGroupSuggestQuery = function() {
      var data = {
        jsonrpc: '2.0',
        method: 'hostgroup.get',
        params: {
          output: ['name'],
		  real_hosts: true,   //Return only host groups that contain hosts
          sortfield: 'name'
        },
        auth: this.auth,
        id: ++this.requestId
      };

      return this.doZabbixAPIRequest(data);
    };

nucleusv:
Added pull request with additonal filters.

Found a lot of bugs, with visualization something strange happens, and there are still errors, will do screenshots.

What about your zabbix, can you applly patch to your zabbix installation.
I will do you it to my. And we can test trends ?

alexanderzobnin:
Please, open issue with screenshots.
Trends is good idea, I'll try to test it later.
Probably, I need to create development roadmap with next goals.

Issue #66, opened 2015-08-12
Issues with Grafana 2.1.1


jbfavre:
Hello,
I've upgraded Grafana in version 2.1.1.
I want to use [Repeating Rows and Panels](http://docs.grafana.org/guides/whats-new-in-v2-1/#repeating-rows-and-panels) feature, amongst others.

But I got some issues (data points not found, panel size, etc...) and wonder if it can come from grafana-zabbix.
Is grafana-zabbix 100% compatbile with grafana 2.1.1 ?
If not, I'm sorry for the noise and will open an issue against grafana.

Thanks,

alexanderzobnin:
Hello. Yes, Grafana-Zabbix 2.0.1 is compatible with Grafana 2.1. I am use Grafana-Zabbix from develop branch with Grafana 2.1.0 at http://play.grafana-zabbix.org. You can check your bugs here.

alexanderzobnin:
Can you give more details about these bugs?

jbfavre:
hmmm... I think I start to understand what happen.
My dashboard is templated with 3 variables:
$group which is static
$POP which extract part of server FQDN. Multi format set to glob
$scenario which is part of items. Multi format set to glob

I then created a graph with a filter in host (dashboard variable called $POP) and another filter on item (dashboard variable called $scenario).
Graph use Repeat Panel based on $scenario.
Row use Repeat Row based on $POP

When I have only one $POP or $scenario selected, that's OK. As soon as I select 2 of each (either $POP or $scenario), I loose data.

I just changed Multi format from "glob" to "regexp value". now it have all my data.
But all data from all POP are on the same graph :(

Comments welcome

jbfavre:
For the "panel size" issue, even if I force panel size with `span 4`, it somewhat looses this value.
So when panel get repeated, new panels get default size which is 12 :-/

alexanderzobnin:
About panel size - it seems like a Grafana bug.

alexanderzobnin:
Templated variables in hosts and item filters works very simply:
```
// Use templated variables in filter
var host_pattern = new RegExp(templateSrv.replace(target.hostFilter));
```
If you use multiple values, templated variable will be transform to something like this:

*glob*
```
{Discovered hosts,Hypervisors,Linux servers,Templates,Virtual machines,Zabbix servers}
```

*regexp value*
```
[...]

PR #67, opened 2015-08-12
Enable scoped variables in Zabbix datasource

jbfavre:
As seen in alexanderzobnin/grafana-zabbix#66, repeat panel and row does not work with Zabbix datasource.
This comes from Host and items filters which does not support it.

I think I found a way to enable them. Not sure it'll cover all cases, but at least my dashboard works now :)

### DIFF
--- a/zabbix/datasource.js
+++ b/zabbix/datasource.js
@@ -102,10 +102,10 @@ function (angular, _, kbn) {
           }
 
           // Replace templated variables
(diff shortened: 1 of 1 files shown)


alexanderzobnin:
@jbfavre Thanks! This is new feature in Grafana 2.1 that I haven't tried.

alexanderzobnin:
Fixed #66

PR #85, opened 2015-09-23
path fix & use dateMath instead of kbn

Sotfo:
@alexanderzobnin with this fix,Grafana-Zabbix would work with Grafana-2.5.


Sotfo:
module path was changed since https://github.com/grafana/grafana/commit/4ffa26cf2cd9ec50a2d8227b953813e9dd5aba41 ,and yes i'll delete `query` and `annotations` from partials.

PR #90, opened 2015-10-05
Adjustments for Grafana 2.5

volter:
Hello!

You might want to add conditionals or add this to a separate branch to keep it working with older version. The two commits contain adapatations for commits done in September.

Unfortunately I found the query editor to be broken in pre-2.5 too, but I have no solution to this yet.

### DIFF
--- a/zabbix/datasource.js
+++ b/zabbix/datasource.js
@@ -2,11 +2,12 @@ define([
   'angular',
   'lodash',
   'kbn',
+  'app/core/utils/datemath',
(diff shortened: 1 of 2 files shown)


alexanderzobnin:
Hello! Thanks for you PR, but this changes also in [fix-query_editor_in_2.5](https://github.com/alexanderzobnin/grafana-zabbix/tree/fix-query_editor_in_2.5) branch. And I fixed query editor for 2.5.

PR #106, opened 2015-10-29
Item select and regex filter with macro replacement

christianchristensen:
See also: https://support.zabbix.com/browse/ZBXNEXT-2841

### DIFF
--- a/zabbix/zabbixAPIWrapper.js
+++ b/zabbix/zabbixAPIWrapper.js
@@ -348,14 +348,19 @@ function (angular, _) {
         // map/index macros by the macro name
         macros = _.indexBy(macros, 'macro');
 
-        // return items replaced by macros
-        var regexp = RegExp ('\\b(' + Object.keys (macros).join ('|') + ')\\b', 'g');
-        return _.each(items, function(i) {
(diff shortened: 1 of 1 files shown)


christianchristensen:
Before:
screenshot 2015-10-29 at 3 24 59 am

After:
screenshot 2015-10-29 at 3 22 34 am

alexanderzobnin:
@christianchristensen Thanks for work! I will try to test your fix soon.

alexanderzobnin:
@christianchristensen Your code works fine with macro but causes a problems with some other metrics. I will try to find a reason of this bug. Try to test with metrics which uses regex filters.

fabpioli:
Good afternoon, Brother ... How can we carry out the implementation in our Gragana the server? I have Grafana 3.1.1 to 3.0.4 zabbix - not have a step by step? I await your reply!

Fábio

alexanderzobnin:
@fabpioli This PR is out of date. But I need to add this feature into the next release.

fabpioli:
Thanks Alex! This topic has been solved with this statement! But it would be very good a new update!

PR #117, opened 2015-12-03
append suffix  "api_jsonrpc.php" to url

WJIAN:
In prxoy model, Grafana alway appends a proxypath to target url, but grafana-zabbix doesn't set it, so I got a url likes "http://xxx.xxx.xxx.xxx/api_jsonrpc.php/", zabbix api return http code 404. I try to fix it with append api_jsonrpc.php as a proxy path.


alexanderzobnin:
@WJIAN What a reson for closing this PR?

PR #147, opened 2016-01-22
Add Debian package source

2ion:
Installation paths are adjusted to match the Debian packages provided by the Grafana upstream project. [Here](https://github.com/2ion/grafana-zabbix/tree/debian) is an orphaned branch with just the Debian files if that is more useful.


alexanderzobnin:
@2ion Grafana is planning to launch plugins marketplace for upcoming Grafana 3.0. And will be very install plugin from it. I think maintaining debian packages not make a sense due this reason. What you think?

2ion:
If the new plugin infrastructure allows for automated installation & updates and removes the need for the manually moving files around, that is if Grafana does the plugin package management, then I would consider this packaging obsolete because I used it to remove the need to do exactly this very same thing when doing multiple deployments.

I didn't know about the plans of the Grafana project; so I would opt to close this now – people who need a package will still be able to find this.

PR #167, opened 2016-03-08
Add zabbix-triggers as metric

steph-ben:
This is my try on how to add zabbix triggers "as metric".
## Goal and achievement

My main goal is to generate a dashboard of trigger status, green if ok, red if not ok. I think with grafana and thanks to this plugin, we can really have the best of the two world !! It will save me a great amount of time, since we are currently making this kind of dashboard with various scripting ...

Here is a screenshot of what we can do with this fork:
image

What do you think about it ?
## Fork status

I'm not familiar with Angular at all ... here is what I didn't succeed to do:
- I beleive there should be another datasource type "Zabbix trigger". However I cannot get it to work, so I modified directly the existing datasource ... Therefore before merging it should eventually create another directory "zabbix-trigger" ?

Minors:
- It looks like the selected trigger is not saved
- I would like to get the trigger description "expanded", ie. no {HOSTNAME} anymore. But then it's hard to get the appropriate events.
- I didn't try with several trigger in the same graph
- I've seen too late there is on-going developement on another branch ...

### DIFF
--- a/zabbix/datasource.js
+++ b/zabbix/datasource.js
@@ -91,7 +91,7 @@ function (angular, _, dateMath) {
      * @return {Object}           Grafana metrics object with timeseries data
      *                            for each target.
      */
-    ZabbixAPIDatasource.prototype.query = function(options) {
+    ZabbixAPIDatasource.prototype.queryTrigger = function(options) {
 
       // get from & to in seconds
       var from = Math.ceil(dateMath.parse(options.range.from) / 1000);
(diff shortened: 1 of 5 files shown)


steph-ben:
Actually, using your latest development branch from http://alpha.grafana-zabbix.org/, I might be able to achieve what I want using the Zabbix trigger panel at it's minimum:
image

Also, I didn't see that we can change metric type from the left menu of metric query. That should be the good way to see "Trigger as metric", if anyone think this is a good idea afterall.

alexanderzobnin:
Thanks for PR, it's good idea! There are many changes in develop branch, I'll try to deal with your PR after 3.0 release.

dizzzz:
any update? there are git conflicts now to be resolved...

steph-ben:
I think there has been many changes in the new version of grafana & grafana-zabbix, so this PR needs to be rewrited quite a lot before being merged.

dizzzz:
ok thnx

evandro-portugal:
Hi guys, I see that this code has been merged. Is this feature working in the current version of the plugin?

evandro-portugal:
This would be a very useful feature to build something like the maps feature in zabbix, it's a great idea, I'll be waiting for it =D

SuperKy3mu4:
I'll be waiting for it too.

evandro-portugal:
Any ETA?

sirmaxru:
I'm using a little workaround to solve this.
You can use IT Service monitoring to monitor trigger status.

PR #207, opened 2016-04-28
feat(query editor): Updated query editor to use new forms styles/markup and outer component

torkelo:
Just merged a breaking change to master (sorry to do this this late in beta phase), but want to update the query editor styles now before more data sources are made, so the breaking change is less painful now. 

this PR updates the zabbix data source query editor. Had to move the menu actions to the actual query editor as the query-row-component that controls the menu and the other actions does not support custom panel actions yet. 

image


alexanderzobnin:
Thanks a lot for your PR, Torkel!

PR #263, opened 2016-08-24
use gray font to distinguish acknowledged triggers

xiaohuiz:
when both acknowledged and un-acknowledged triggers are displayed, it is not very clear to see the difference. 
this patch gray out the acknowledged ones to enable end users ignore them easily.

### DIFF
--- a/src/panel-triggers/module.html
+++ b/src/panel-triggers/module.html
@@ -46,7 +46,7 @@
               {{trigger.severity}}
             </div>
           </td>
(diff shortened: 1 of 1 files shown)


alexanderzobnin:
@xiaohuiz Can you attach some screenshots?

xiaohuiz:
as in the picture, with this patch, on can easy say the first one is acknowledged(somebody has already take care of it) and others not:
image

alexanderzobnin:
@xiaohuiz Looks good

PR #268, opened 2016-08-30
allow regex in template query

xiaohuiz:
treat each part of the template query as regex to enable: 
1. more flexible group/host/app/item filter
1. using Tags in template variables


xiaohuiz:
some screen shoots:

this query select all hosts with a number in its name:
image

Tags query: select host groups Network and Linux
Tag values query: select hosts in that groups
image

Tags displayed in the variable drop down box:
image

Issue #271, opened 2016-09-05
Text items not including hostname


potchin:
I'm not sure if this is a zabbix or zabbix-grafana issue but when querying text metrics into a table for more than 1 server it seems the hostname is not included in the data.

In my case I am querying for a package version info for a bunch of servers. I get the versions back but I have no idea which version corresponds to which server.

alexanderzobnin:
@potchin I understand. I'm going to add table format support for Zabbix response and I think it's possible to add some info to result, such as host name, item name, may be some additional fields.

tinohean:
how is this issue going?  any support for text metrics now? thanks a lot ~

PR #272, opened 2016-09-06
Display host column in table panel

tgrperf:
I need to get the same metric from many hosts on one table panel.
So i have to display the host column, in table panel with timeseries to rows model.
is that possible?? zabbix datasource is used for this table panel.

Thanks in advance.


alexanderzobnin:
@tgrperf see #271

Issue #205, opened 2016-04-27
Template variables not resolved in functions.


andrewmcgilvray:
I am using a template variable that resolves correctly in the Item field but when I try to use it in setAlias() it does not resolve. It does get coloured light blue in the editor as a known variable but does not get replaced in the Legend.

alexanderzobnin:
Yes, I need to implement it.

2ion:
Imported from #217.
> I am in the process of migrating a Grafana 2.6 dashboard using grafana-zabbix to 3.x and am noticing a loss of flexibility in the handling of query aliases which is useful in particular when it comes to Zabbix items that have very long names and using template variables to select Zabbix hosts.
> 
> While when using a template variable and a query alias in 2.6 the resulting item name was a concatenation of the "$hostname: $itemname" and only $itemname was replaced with the query alias, in Grafana 3 and latest grafana-zabbix when using the function setAlias($alias) to set an alias instead, the resulting item name is just $alias.
> 
> This makes the feature less useful when used in conjunction with template variables: Previously, you could go and have i.e. a load average graph for 50 hosts with data series labels reading "$hostname:$itemname" where $itemname was empty because we replaced it by setting an alias. After all, it doesn't make sense to repeat 50 times "$hostname: Processor load (5 min average per core)" in the graph legend - putting that into the graph title does. Because of the new behaviour of the setAlias() function, in Grafana 3, we can't have $hostname in the legend anymore when using the same hostname templating technique.
> 
> In order to be able to do the same in Grafana 3 with grafana-zabbix, please expand template variables used as the function argument to setAlias(). As of this moment, the variables are not expanded but printed literally to the UI.

dbason:
Template variable would be useful in other functions asides from the setAlias as well.  For example using an interval template with groupBy would allow me to smooth the graphs on larger time periods.

alexanderzobnin:
Yes, it would be useful.

2ion:
@alexanderzobnin Thank you very much for resolving the issue; however even with this the application as described in my previous post can't be implemented correctly.

[...]

PR #273, opened 2016-09-07
Show only hostnames as labels when multiple hosts were selected

2ion:
...while keeping setAlias() working. The variable system was hard to debug and I didn't have time to spend on studying Grafana in detail.

This is my final solution for my problems described in #205. I'm opening this just as a suggestion/proof of desireability, please feel free to close this PR if you disagree with this way of doing things or see usablity problems (I do, e.g. when using different kinds of metrics in the same graph).


alexanderzobnin:
@2ion ok, thanks!

PR #284, opened 2016-09-28
new function percentil

pedrohrf:
hi, due to a need for the company where I work, I created a new percentile function in your project, the purpose of it is to make a kind of median, only instead of always taking the middle value (50%), it will take any position that is passed (n%). In my language is called "percentil" I do not know if in English is how it is written. It would be interesting for us the project stay synchronized with yours development, in order to use your future updates.

### DIFF
--- a/src/datasource-zabbix/DataProcessor.js
+++ b/src/datasource-zabbix/DataProcessor.js
@@ -158,6 +158,32 @@ export default class DataProcessor {
     return timeseries;
   }
 
+  static PERCENTIL(interval, n, datapoints) {
+    var flattenedPoints = _.flatten(datapoints, true);
+    var ms_interval = utils.parseInterval(interval);
+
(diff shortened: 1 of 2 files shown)


alexanderzobnin:
@pedrohrf Thanks a lot. In english it's called [Percentile](https://en.wikipedia.org/wiki/Percentile). But nevermind, I'll refactor that after merging.

andreioss:
Alexander, it would be great, if you merge this feature, because percentile is really needed function for a lot of graphs. Could we expect it in the nearest future?

joriws:
Is there any update. I've the latest published zabbix plugin via grafana-cli and it does not have percentile-function. Naturally I could apply pedrohrf's patch by hand but I'd really like to see this baked in to official plugin.

igorfernandes:
Any news about this function?

PR #386, opened 2017-04-25
Update README.md

Laakso:
Updated year from 2016 to 2017.


alexanderzobnin:
@Laakso ah, sorry, i've made the same change, just haven't pushed it.

PR #427, opened 2017-07-14
Fix "data points outside time range" when there is no datapoints and aggregate function is used

ma-tty:
Patch to fix misleading "data points outside time range" alert message on a graph using aggregate function (say median(10m)) if metric has no data in a selected time range.
groupBy_perf function in datasource-zabbix/timeseries.js do not properly handle empty datapoints[] case inserting bogus [undefined, 0] sample with last grouped_series.push([frame_value, frame_ts]) call.
image
image


alexanderzobnin:
I'll check it soon, thanks for PR!

Issue #352, opened 2017-02-09
Application selection in metric does not work


Tony-Mc:
- What Grafana version are you using?
v4.1.1

- What Zabbix version are you using?
v2.0.6

- What zabbix plugin version are you using?
v3.2.0

- What OS are you running grafana on?
[Docker Grafana image](https://hub.docker.com/r/grafana/grafana/)

- What did you do?
I set up a metric for application called "MyApp".
![image](https://cloud.githubusercontent.com/assets/8818187/22787003/6fb640e8-eee3-11e6-945c-3c024c08ee90.png)

- What was the expected result?
It should have displayed metric for selected application ("MyApp").

- What happened instead?
Stats for that metric were not displayed because application was not found (I also tried regexp instead of name).

As far as I debugged this issue the problem is that Zabbix datasource does not include `expandData` flag in `application.get` [request](https://github.com/alexanderzobnin/grafana-zabbix/blob/master/src/datasource-zabbix/zabbixAPI.service.js#L141) for Zabbix versions [2.0](https://www.zabbix.com/documentation/2.0/manual/appendix/api/application/get)-[2.2](https://www.zabbix.com/documentation/2.2/manual/api/reference/application/get). Due to that `name` field is not returned, making search by name or regexp impossible.

alexanderzobnin:
Fixed by #462

PR #462, opened 2017-09-15
Request extended output for applications from service

Tony-Mc:
Solves issue #352
However, I've only tested it on Zabbix 2.0.6.
According to it's docs `output: 'extend'` should not break anything for other versions, but I might be wrong.


Tony-Mc:
Hi @alexanderzobnin, could you review this?

PR #469, opened 2017-10-23
Fix threshold regex for equality expressions

akotynski:
This PR improves the retrieval of alert thresholds with the expressions `<=` `> =` `=`.

Btw, I upgraded jsdom to newest version so contextify is no longer downloaded as dependency (the requirement for installed g++ was very painful)
I created .travis.yml so you should enable automatic builds :)


alexanderzobnin:
That's very cool! I'll try to test it and merge ASAP.

Issue #471, opened 2017-10-23
Set single stat thresholds from triggers 


akotynski:
### I'm submitting a ...
- [ ] Bug report
- [x] Feature requestdd
- [ ] Question / Support request

Recently I created dashboard of single stats using grafana template. Each item has triggers configured in Zabbix. It would be nice to provide functionality to set thresholds from zabbix triggers.  I already prepared Proof of Concept and it almost work - proper background color is set after first dashboard refresh (maybe you could help me with that? I need a way to re-render dashboard).

Possibilities:
- always replace thresholds when triggers are available
- replace thresholds on checked checkbox in configuration
- replace thresholds on keyword ex. `auto` in field thresholds (I'm not sure if this is possible)

In grafana we have 3 states of singlestat (ok, warn, error) but in Zabbix we can define much more triggers. My proposition is to set first threshold value from first warning/average trigger and second threshold from first high/disaster trigger. In simplest solution we can take first and last value from array of triggers (I think that in most cases people have only warning and error triggers).

Tomorrow I can provide you code samples.

alexanderzobnin:
Ok. I change thresholds for graphs and this should be possible for singlestat as well.

alexanderzobnin:
I'm thinking about this feature and actually can't find a good solution. I don't like the fact that query result makes changes in the dashboard model. And because of this I  don't like how this feature works now for graph panels. Dashboard model should be intact. So I want to find another way to implement this. Maybe, template variables are suitable for this, not sure.
@akotynski any thoughts?

PR #473, opened 2017-10-25
add thresholds to singlestat and alarmbox plugins

akotynski:
PR for request #471  
~~There is an issue which I described earlier. When on dashboard I have 40 items, not every item has proper background color.
Look at last item witch has active trigger:
grafana_invalid
After refreshing dashboard with dedicated grafana button bg colors are correct:
grafana_after_refresh
I'm not sure how to deal with this problem.~~

Btw. I think that you should add folder dist/test to gitignore (or even whole dist folder...).


akotynski:
You should enable travis build since you have `.travis.yml` in your repo

alexanderzobnin:
Yea, will configure builds soon,

akotynski:
Ok, I handled issue  I mentioned before. Check last commit. Problem was with ignoring result of alertQuery. Now we're waiting for results and everything seems ok!

akotynski:
I added new scoped variables: dashboard and thresholds. Usage example on image:
grafana_zabbix

alexanderzobnin:
I can't add whole `dist/` to gitingnore, because this required by grafana (this dir contains builded plugin), but I think I can add `dist/test`.

akotynski:
```The plugin.json file is mandatory for all plugins. When Grafana starts it will scan the plugin folders and mount every folder that contains a plugin.json file unless the folder contains a subfolder named dist. In that case grafana will mount the dist folder instead.```
http://docs.grafana.org/plugins/developing/plugin.json/
I think that you should create package only with mandatory files as travis build result. In this case you could add whole dist folder to .gitignore ;)
Tomorrow I will resolve conflicts.

**Comment by alexanderzobnin** [comment]
@akotynski I did this earlier, but there're some pros and cons of this approach. I should maintain 2 repos - one with sources only and another - with builded plugin (because `grafana.com/plugins` and `grafana-cli` uses git repo address). It may be easier with CI system to do that, I should think about it, because I also don't like a thing I have to store built plugin in the same repo.

File: src/datasource-zabbix/datasource.js [hunk 1]
```diff
@@ -148,11 +147,11 @@ class ZabbixAPIDatasource {
     });
 
     // Data for panel (all targets)
-    return Promise.all(_.flatten(promises))
+    return emptyPromise.then(() => Promise.all(_.flatten(promises))
```
alexanderzobnin: Why do we have to wait for `alertQuery()` response?

src/datasource-zabbix/datasource.js
```diff
@@ -640,6 +640,15 @@ class ZabbixAPIDatasource {
     );
     return useTrends;
   }
+
+  setDescription(items, options) {
+    let panelModel = this.dashboardSrv.dash.getPanelById(options.panelId);
+    let desc = items.map(i => i.description);
+    desc = desc.length === 1 ? desc[0] : desc;
```
alexanderzobnin: Cant understand the logic of this. `desc` is array, why should we assign it to variable's value?

src/datasource-zabbix/datasource.js
```diff
@@ -640,6 +640,15 @@ class ZabbixAPIDatasource {
...
+    panelModel.scopedVars.description = {
```
alexanderzobnin: This cause error when `panelModel.scopedVars` is undefined. You should set it to default value before: 
```js
panelModel.scopedVars = panelModel.scopedVars || {}
```

alexanderzobnin:
@akotynski Sorry for delay, hadn't enough time for review

src/datasource-zabbix/datasource.js
akotynski: Look at the screenshot in description. Without that some alerts had wrong background color because thresholds are set while fetching alerts. 
At first screenshot the last alert has red heart but green background. It should be orange like on second screenshot.

src/datasource-zabbix/datasource.js
akotynski: Array will produce string: `[ item ]` but single value just `item`. So, when there is only one item I decided to unwrap this item from array.

src/datasource-zabbix/datasource.js
akotynski: ok

alexanderzobnin:
@akotynski I added my thoughts to the original issue.

PR #526, opened 2018-02-14
Additional options for Format as table feature

akotynski:
I found your latest feature `format as table`. It's really nice but I need some customization. I'd like to have only the `Host` and `Last Value` columns (without `key` and `item`).
In addition, I do not want items with an empty value to be visible.

I prepared PR with additional options for displaying the table. By default all columns are checked and `skip empty values` is unchecked.
table_customization
With this options I could create monitoring like this:
table_example

When everything is ok, there are empty values so my list is empty ;)


alexanderzobnin:
In Grafana table panel you can hide columns by using type _Hidden_. Is it good enough to don't duplicate this in plugin? _skip empty values_ may be really helpful.

akotynski:
You're right. I couldn't find this option in grafana...

PR #576, opened 2018-05-20
Add placeholder for suggest-url

mammuthus:
Because it was empty.
Moreover, it requires not usual url, but url to Zabbix Api file (api_jsonrpc.php), so It it will be easier to connect with this placeholder.

Before:
before

After:
after


alexanderzobnin:
Yes, this makes sense.

PR #596, opened 2018-07-11
Links to zabbix server

belkir:
added links to zabbix servers
like this:
`<span ng-if="trigger.datasource_url" class="zabbix-trigger-name"><a ng-href="{{trigger.datasource_url}}" target="_blank" title="Open in Zabbix">{{trigger.description}}</a></span>`


alexanderzobnin:
I don't think this should be merged. What's a problem this PR solves? Is there an issue created for this problem?

Issue #603, opened 2018-07-26
consolidateBy(sum) does not work correctly (Direct DB Connection)


kompa3:
### I'm submitting a ...
- [ X ] Bug report
- [ ] Feature requestdd
- [ ] Question / Support request

### For bug report please include this information:

**Software Versions**

| Grafana | Zabbix | Grafana-Zabbix Plugin |
| ------- | ------ | --------------------- |
| 4.6.2   | 3.4.3  | 3.9.0                 |

#### What did you do?

The function consolidateBy(sum) should calculate sum over the data points returned from Zabbix database to Grafana. It works well when querying from history tables (time range being less than the Range setting in plugin settings) but calculates invalid values when querying from trends (from longer time ranges).

![image](https://user-images.githubusercontent.com/9412499/43259846-5b404c8a-90e0-11e8-93ce-6940ddd6c5f4.png)

The produced SQL query can be debugged in network traffic:

```
SELECT itemid AS metric, clock DIV 7200 * 7200 AS time_sec, SUM(value_avg) AS value FROM trends_uint WHERE itemid IN (282744, 282835) AND clock > 1531999580 AND clock < 1532604380 GROUP BY clock DIV 7200 * 7200, metric ORDER BY time_sec ASC
```

The query shows that Zabbix plugin actually sums up the average values from trends_uint table, meaning that it sums up one-hour averages instead of summing up the actual data inside those one-hour trend periods.

Zabbix trend tables do have the capability to get the sum of actual data inside the one-hour period. Zabbix stores the number of datapoints inside the one-hour trend period into `num` field and the average value in `value_avg` field, meaning that the sum of values inside the one-hour trend period can be acquired by `num*value_avg`.

PR #604, opened 2018-07-26
consolidateBy(sum) must calculate sums inside the one-hour trend period

kompa3:
This fixes https://github.com/alexanderzobnin/grafana-zabbix/issues/603.

After the fix, the produced SQL query looks like the following:
```
SELECT itemid AS metric, clock DIV 7200 * 7200 AS time_sec, SUM(num*value_avg) AS value FROM trends_uint WHERE itemid IN (282744, 282835) AND clock > 1531999580 AND clock < 1532604380 GROUP BY clock DIV 7200 * 7200, metric ORDER BY time_sec ASC
```

The calculated sum is correct since it sums the actual data values and not the one-hour trend period averages.


alexanderzobnin:
@kompa3 could you please rebase this?

kompa3:
> @kompa3 could you please rebase this?

Rebased

Issue #387, opened 2017-04-25
Arithmetic function feature request


Laakso:
[x ] Feature request

Would be nice to have some sort of function which can do basic arithmetic calculations such as subtraction. Use case for my need is: I'm querying CPU Idle % from Zabbix to Grafana. I would like to graph actual CPU utilization like (100 - CPU Idle %). What do you think?

alexanderzobnin:
I also have multiply operation (`scale` function), so it makes sense I think.

gelonsoft:
+1

RedHat93:
+1 will be really useful

drakosha:
What functions are needed except shift, that i already added in #637?

PR #638, opened 2018-10-30
Offset function added

### LINKED ISSUE #637: Add shift function to metrics
Usually zabbix have next CPU metrics: idle, system, user, iowait, nice. But, we haven't 'utilization' metric. The simple way to have it, it's scale idle by -1. The problem is that in this case you will have negative values and there are no way to shift them by 100.

So, we need a shift function for metric.

drakosha:
Will resolve #637.

Adding a shift function.

### DIFF
--- a/src/datasource-zabbix/dataProcessor.js
+++ b/src/datasource-zabbix/dataProcessor.js
@@ -9,6 +9,7 @@ let sumSeries = ts.sumSeries;
 let delta = ts.delta;
 let rate = ts.rate;
 let scale = (factor, datapoints) => ts.scale_perf(datapoints, factor);
+let shift = (factor, datapoints) => ts.shift(datapoints, factor);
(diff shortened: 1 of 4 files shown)


Laakso:
Yes finally. 😊

alexanderzobnin:
@drakosha thanks for the PR! Could you rename function from `shift` to `offset`? My motivation is to keep names accordingly to other Graphite functions, like `scale`, `sumSeries`, `timeShift`, `movingAverage` and other. There's [`offset`](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.offset) function in Graphite, but no `shift`. Also, `shift` may be confused with `timeShift`.
This will fix #387.

drakosha:
@alexanderzobnin not a problem, done.

Issue #752, opened 2019-06-05
Consolidated average is not accurate with direct DB connection


kompa3:
**Describe the bug**
When querying metrics from direct DB connection with the default consolidateBy(avg), the resulting average metrics is not exactly correct. The issue is most clearly seen when the number of queried data points is small.

**Expected behavior**
The calculated average should be the same as calculated from database without any grouping, ie just a simple avg() over the time period.

**Underlying issue**
SQL query uses grouping by clock integer division (DIV). However, with small data point number and consequently big intervals, the integer division is arbitrarily placed along the time period. When the average is calculated from the resulting data points, it is not correctly weighted.

Simplistically illustrated:

* Interval is 30 days and the query is made of a calendar month
* The integral division, 30 days, may change let us say at the 10th day of the month
* Query results two data points: average of 1..10th days and average of 11th..30th days
* The average from the data points is weighted to favor the latter part of the month because it does not calculate a weighted average but a basic average of the data points

**Software versions**

| Grafana | Zabbix | Grafana-Zabbix Plugin |
| ------- | ------ | --------------------- |
| 5.4.3   | 4.0.5 | 3.10.2 |

PR #753, opened 2019-06-05
Fix to consolidate avg accuracy

kompa3:
Fix for issue https://github.com/alexanderzobnin/grafana-zabbix/issues/752


kompa3:
I have tested the effects of this fix and it seems to produce more exact results now. There still remains cases where the values are not evenly distributed along the timeline, causing the downsampled average of the database-time-based-group-averages being not equal to the average calculated from the database values directly.

Issue #190, opened 2016-04-08
Trigger Panel Info Column URL with Macros not resolving


mdiorio:
This seems to be a Zabbix issue but wanted to make you aware of it.

So my Triggers for website monitoring are all templated and working properly.  I have a URL on the trigger with a Macro in it and it works fine within Zabbix.  I can view the trigger and open the URL and it processes the macro

http://{HOST.HOST}/content-check/checknow.cfm?mode=all

The Trigger Panel is not getting the macro converted.  I'm getting an URL encoded, un-converted macro

http://%7Bhost.host%7D/content-check/checknow.cfm?mode=all

In the API, there are flags for:
expandComment   flag    Expand macros in the trigger comment.
expandDescription   flag    Expand macros in the name of the trigger.
expandExpression    flag    Expand functions and macros in the trigger expression.

But there is no equivalent for expandURL, so as you can see, macros don't get expanded when requested through the API.

I cross posted in the Zabbix forum as well:
[https://www.zabbix.com/forum/showthread.php?p=181416#post181416](url)

Hope this gets implemented on the Zabbix side quickly.

Thanks.

Max

volter:
Which version of Zabbix is that?

mchihi:
any update please regarding this issue. using Macros instead of host name.
http://{HOST.HOST}/content-check/checknow.cfm?mode=all

alexanderzobnin:
This is Zabbix API limitations.

timyllym:
I have the same problem. I think there won't be any fixes to Zabbix API in near future. I fixed this by utilizing existing code and made a PR #759.

alexanderzobnin:
@timyllym Thank! I'll take a look.

PR #759, opened 2019-06-19
Expand macros in trigger URL fields

timyllym:
Fix to issue #190 

Currently macros in trigger URL field don't get expanded.
This fix checks trigger URL against macros and expands all found macros.


duuz:
+1

Issue #728, opened 2019-04-23
Feature request: SLA graph over time


mjtrangoni:
Hi @alexanderzobnin,

**Is your feature request related to a problem? Please describe.**
The SLAs are available via [service.getsla](https://www.zabbix.com/documentation/4.2/manual/api/reference/service/getsla), which gives us a Graph with a continuous line from timestamp A to timestamp B.

**Describe the solution you'd like**
I would like to have a variable graph, that change the slope when the SLAs changes inside the timestamp period.

**Describe alternatives you've considered**
I see, there is a **Service availability report** with the option "daily", which would get us at least a daily resolution for this, for example here,
```
https://zabbixurl/zabbix.php?action=report.services&serviceid=1&period=daily&year=2019
```

**Additional context**
I am new to Zabbix, so I don't known the background, why this is not implemented over the time already. I would like to hear your opinion about it, and if this should be directly a change request at the zabbix-side.

A kind of workaround/solution would be, sending more than only one interval with `service.getsla`, so that you would get more values back.

flassman:
Another workaround would be to query the SLA values from zabbix itself (with a web scenario?) to get the usual data which can be nicely displayed like any other graph

alexanderzobnin:
That's tricky and may cause Zabbix server overloading. I strongly believe that if you want to graph something, so a corresponding method for this should exist.

SeverinLeonhardt:
According to the [service.getsla documentation](https://www.zabbix.com/documentation/4.2/manual/api/reference/service/getsla) the `intervals` parameter accepts an array of intervals. This should allow querying multiple SLA values within one API request to generate a graph.

alexanderzobnin:
@SeverinLeonhardt yes, I understand. But how many intervals I can provide without affecting zabbix performance. I see that in reports Zabbix uses weekly intervals only.

JeferCatarina:
Hi @alexanderzobnin how've you been? 

My company was looking for this feature in Grafana and I made some tests with your code and I think that could be useful in general. My idea was keep the same interval as in the Zabbix PHP inside the Grafana plugin, that is 1 day. With this interval I think it wouldn't affect performance and we could summarize data using the functions. E.g: An interval of 7 days will return 7 results. To show the SLA over time, just plot this values in a "Graph" and to show a summarized SLA plot this values using "Single stat" with "avg" or adding the function "aggregateBy" in a "Graph".

What do you thing? Could be a new feature?

[...]

PR #764, opened 2019-07-01
Feature 728 sla graph over time

JeferCatarina:
PR related: #728 

Added support to show IT Services graph over time and to apply "Functions" on it. 
* Interval between each SLA range: 86400s (1d);
* When the time range is less than 1d it will return the exactly range specified;

image

image


alexanderzobnin:
@JeferCatarina Thanks for the contribution! I think the same intervals as Zabbix provides makes sense. PR looks good overall, I need to do a more detailed review when I have time.

Issue #802, opened 2019-09-26
Backend: cache for Zabbix API requests


alexanderzobnin:
Zabbix plugin produces a lot of queries, so in order to rid of unnecessary workload, cache should be implemented on the backend. [go-cache](https://github.com/patrickmn/go-cache) can be used as a simple cache implementation.

See `cachingProxy.js` for details
https://github.com/alexanderzobnin/grafana-zabbix/blob/6e80b491205168a2744970bfd0403a76447e580b/src/datasource-zabbix/zabbix/proxy/cachingProxy.js#L70

PR #807, opened 2019-10-03
Backend: caching the Zabbix API Queries

vignesh-reddy:
Issue https://github.com/alexanderzobnin/grafana-zabbix/issues/802

Implemented caching for API Queries by using go-cache as mentioned in the issue. Converted the DatasourceRequest to a hash string and stored the response against it. The expiration time and cleanup time are given as 10 minutes.


alexanderzobnin:
Great! I'll take a look.

Issue #799, opened 2019-09-26
Backend: renew auth token if expired


alexanderzobnin:
Backend should check if auth token is expired and obtain new in this case. This can be done by checking zabbix api response for error in the same way as frontend code does

https://github.com/alexanderzobnin/grafana-zabbix/blob/6e80b491205168a2744970bfd0403a76447e580b/pkg/zabbix_api.go#L116-L121

https://github.com/alexanderzobnin/grafana-zabbix/blob/6e80b491205168a2744970bfd0403a76447e580b/src/datasource-zabbix/zabbix/connectors/zabbix_api/zabbixAPIConnector.js#L48-L58

PR #808, opened 2019-10-03
Backend: Added logic to renew auth token if expired

vignesh-reddy:
issue https://github.com/alexanderzobnin/grafana-zabbix/issues/799

Added logic to renew the auth token if expired and the default number of attempts allowed is set to 3.

### DIFF
--- a/pkg/zabbix_api.go
+++ b/pkg/zabbix_api.go
@@ -116,15 +116,18 @@ func (ds *ZabbixDatasource) ZabbixRequest(ctx context.Context, dsInfo *datasourc
 	var err error
 
 	for attempt := 0; attempt <= 3; attempt++ {
-		// Authenticate
-		auth, err := ds.loginWithDs(ctx, dsInfo)
-		if err != nil {
-			return nil, err
+		if zabbixAuth == "" {
+			// Authenticate
+			zabbixAuth, err = ds.loginWithDs(ctx, dsInfo)
+			if err != nil {
+				return nil, err
+			}
 		}
-		zabbixAuth = auth
 		result, err = ds.zabbixAPIRequest(ctx, zabbixUrl, method, params, zabbixAuth)
 		if err == nil || (err != nil && !isNotAuthorized(err.Error())) {
 			break
+		} else {
+			zabbixAuth = ""
 		}
 	}
 	return result, err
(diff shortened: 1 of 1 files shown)


pkg/zabbix_api.go
```diff
@@ -112,16 +112,22 @@ func (ds *ZabbixDatasource) BuildResponse(result *simplejson.Json) (*datasource.
...
+
+	for attempt := 0; attempt <= 3; attempt++ {
+		// Authenticate
 		auth, err := ds.loginWithDs(ctx, dsInfo)
 		if err != nil {
 			return nil, err
 		}
 		zabbixAuth = auth
+		result, err = ds.zabbixAPIRequest(ctx, zabbixUrl, method, params, zabbixAuth)
+		if !isNotAuthorized(err.Error()) {
```
alexanderzobnin: `err` is always `nil` at this point, so it causes a panic

pkg/zabbix_api.go
```diff
@@ -112,16 +112,22 @@ func (ds *ZabbixDatasource) BuildResponse(result *simplejson.Json) (*datasource.
...
+	var result *simplejson.Json
+	var err error
+
+	for attempt := 0; attempt <= 3; attempt++ {
+		// Authenticate
 		auth, err := ds.loginWithDs(ctx, dsInfo)
 		if err != nil {
 			return nil, err
 		}
 		zabbixAuth = auth
```
alecxvs: This appears to be stored at the package level. If I understand the backend API correctly, we may be handling requests from multiple different Zabbix datasources during this runtime though. @alexanderzobnin

alexanderzobnin: @alecxvs yes, you're right. This should be stored in datasource-specific object. I think we should create a cache instance for each datasource.

alexanderzobnin: I think we can leave it as is for now, but I'll create an issue for moving things into datasource-specific cache.

pkg/zabbix_api.go
vignesh-reddy: Updated 4f5f5ec

pkg/zabbix_api.go
alexanderzobnin: @alecxvs I created an issue #812

pkg/zabbix_api.go
```diff
@@ -112,16 +112,22 @@ func (ds *ZabbixDatasource) BuildResponse(result *simplejson.Json) (*datasource.
...
 	zabbixUrl := dsInfo.GetUrl()
 
-	// Authenticate first
-	if zabbixAuth == "" {
+	var result *simplejson.Json
+	var err error
+
+	for attempt := 0; attempt <= 3; attempt++ {
+		// Authenticate
 		auth, err := ds.loginWithDs(ctx, dsInfo)
```
alexanderzobnin: In this case, data source tries to login at every request. We should use stored `zabbixAuth` token and renew it only if "not authorized" error happens.

vignesh-reddy: Good catch, Updated https://github.com/alexanderzobnin/grafana-zabbix/pull/808/commits/371c68dd2e20fe720f565785000145f09ec6f539

Issue #800, opened 2019-09-26
Store password in SecureJsonData


alexanderzobnin:
Since auth is moving to the backend, password should be stored in `SecureJsonData` and shouldn't be exposed to frontend anymore. Related to #312 #380 #684

PR #815, opened 2019-10-10
Backend: Check secure JSON data for password before plaintext JSON data

alecxvs:
Related to #800 

Not sure if Grafana handles all of the datasource configuration, or some part of the frontend needs to be updated to save to the secure data. Also can't currently test this change in Grafana.

Since the change was pretty simple, I made some additional changes to appease golint and cleaned up redundant code that was being used, presumably, for debug logging.

### DIFF
--- a/pkg/zabbix_api.go
+++ b/pkg/zabbix_api.go
@@ -48,16 +48,6 @@ func (ds *ZabbixDatasource) ZabbixAPIQuery(ctx context.Context, tsdbReq *datasou
 		return nil, err
 	}
 
-	jsonDataStr := dsInfo.GetJsonData()
-	jsonData, err := simplejson.NewJson([]byte(jsonDataStr))
-	if err != nil {
-		return nil, err
-	}
-
-	zabbixLogin := jsonData.Get("username").MustString()
-	// zabbixPassword := jsonData.Get("password").MustString()
-	ds.logger.Debug("ZabbixAPIQuery", "url", zabbixUrl, "user", zabbixLogin)
-
 	jsonQueries := make([]*simplejson.Json, 0)
 	for _, query := range tsdbReq.Queries {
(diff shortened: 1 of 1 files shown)


pkg/zabbix_api.go
```diff
@@ -145,37 +130,43 @@ func (ds *ZabbixDatasource) loginWithDs(ctx context.Context, dsInfo *datasource.
...
-	zabbixPassword := jsonData.Get("password").MustString()
-	auth, err := ds.login(ctx, zabbixUrlStr, zabbixLogin, zabbixPassword)
+	var zabbixPassword string
+	if securePassword, exists := dsInfo.GetDecryptedSecureJsonData()["password"]; exists {
+		zabbixPassword = securePassword
+	} else {
+		zabbixPassword = jsonData.Get("password").MustString()
+	}
+
+	auth, err := ds.login(ctx, zabbixURLStr, zabbixLogin, zabbixPassword)
```
alecxvs: This is the change to use secure JSON data.

pkg/zabbix_api.go
```diff
@@ -86,7 +71,7 @@ func (ds *ZabbixDatasource) ZabbixAPIQuery(ctx context.Context, tsdbReq *datasou
 		apiMethod := jsonQuery.Get("method").MustString()
 		apiParams := jsonQuery.Get("params")
 
-		result, err = ds.ZabbixRequest(ctx, dsInfo, apiMethod, apiParams)
+		result, err := ds.ZabbixRequest(ctx, dsInfo, apiMethod, apiParams)
```
alexanderzobnin: This change causes plugin crash because `result` is `nil`. If you use `:=`, `result` will be a new variable in `if {` scope and original `result` will be empty. I fixed it.

PR #816, opened 2019-10-10
Backend: Makefile/CircleCI improvements for Golang

alecxvs:
Proposing changes:
- Handle both Go and Node.js install/build steps in the Makefile to make it easier to run everything together
- Use a vendor directory for Go dependencies so they can be cached during CI builds
- Build backend binaries for Windows and OS X when building for distribution
- Updated plugin.json to configure this as a backend plugin
- Changed the name of a symbol to fix a codespell error
- Changed CircleCI config to use the Makefile
- Switched CircleCI jobs to run Golang + Node.js
- Added Go vendor directory to CircleCI cache
- Switched from codecov-node to codecov-bash in the CI test so that Go coverage could be included (there's another way, using a CircleCI orb, but I'm not sure if that's supported in the current CI config)


src/plugin.json
```diff
@@ -3,6 +3,11 @@
   "name": "Zabbix",
   "id": "alexanderzobnin-zabbix-app",
 
+  "metrics": true,
```
alexanderzobnin: We don't need this here. That options only for datasource `plugin.json`

src/plugin.json
```diff
@@ -26,8 +31,8 @@
       {"name": "Metric Editor", "path": "img/screenshot-metric_editor.png"},
       {"name": "Triggers", "path": "img/screenshot-triggers.png"}
     ],
-    "version": "3.10.4",
-    "updated": "2019-08-08"
+    "version": "4.0.0",
```
alexanderzobnin: I'd prefer to use 4.0.0-beta or alpha at this step

PR #822, opened 2019-10-18
Backend: Handle multiple datasources + caching + connection test

### LINKED ISSUE #812: Per-datasource cache and auth storage
Initial work is done, but as mentioned in https://github.com/alexanderzobnin/grafana-zabbix/pull/808#discussion_r331600052, we should store cache and auth for each particular data source, not for the package at all. So we need an object that stores information related to data source, like:
```golang
type ZabbixDatasource struct {
	plugin.NetRPCUnsupportedPlugin
	logger    hclog.Logger
	instances map[int64]ZabbixInstance
}

type ZabbixInstance struct {
	dsInfo *datasource.DatasourceInfo
	cache  *Cache
	auth   string
}
```
This is just an example, so maybe there's better solution.

alecxvs:
Fixes #812 

Many changes:
- Cache Datasources based on hash of the datasource info
- Move Zabbix API functions to per-datasource type. This should solve having multiple Zabbix datasources
- Implement connection test on the backend + changes to the frontend to accomodate
- Additional changes to the frontend to make object types more obvious
- Upgraded Babel and other dependencies

I've implemented the connection test query type as a means of testing multiple datasources. I haven't fully tested using multiple datasources but the connection test is successful with the datasource caching so that's pretty good for now.

I have a changeset that adds more type definitions and converts jscs/jshint to eslint but I'll open another PR for that later.


src/datasource-zabbix/zabbix/zabbix.js
```diff
@@ -110,37 +110,37 @@ export class Zabbix {
...
-      if (testResult) {
-        dbConnectorStatus = {
-          dsType: this.dbConnector.datasourceTypeName,
-          dsName: this.dbConnector.datasourceName
-        };
-      }
-      return { zabbixVersion, dbConnectorStatus };
-    });
-  }
+  // testDataSource() {
```
alexanderzobnin: We need `testDataSource()` method implemented on the backend side.

alecxvs: There is a datasource test implemented on the backend. I changed the frontend testDatasource to post to the backend here:
- https://github.com/alexanderzobnin/grafana-zabbix/pull/822/files#diff-4e2b5df346a2dd67cd90176dfb35ade5R421
- https://github.com/alexanderzobnin/grafana-zabbix/pull/822/files#diff-4e2b5df346a2dd67cd90176dfb35ade5R208-R223

and the backend is implemented here:
- https://github.com/alexanderzobnin/grafana-zabbix/pull/822/files#diff-2cb344693fe3302e52c776fc8f6e3964R41-R42
- https://github.com/alexanderzobnin/grafana-zabbix/pull/822/files#diff-5641cbffc2e0b7f7310a3c757ca99c6bR117

Unless you mean we specifically need a method called testDataSource in the backend?

PR #844, opened 2019-12-13
Backend: Unit Testing

vignesh-reddy:
This PR contains unit tests for the backend. Considering that all the backend is in single package "main" this PR improves the unit test coverage to 80%

image

### DIFF
--- a/pkg/zabbix_api.go
+++ b/pkg/zabbix_api.go
@@ -105,14 +105,14 @@ func (ds *ZabbixDatasource) TestConnection(ctx context.Context, tsdbReq *datasou
 
 	auth, err := ds.loginWithDs(ctx, dsInfo)
 	if err != nil {
-		return BuildErrorResponse(fmt.Errorf("Authentication failed: %s", err)), nil
+		return BuildErrorResponse(fmt.Errorf("Authentication failed: %w", err)), nil
 	}
 	ds.authToken = auth
 
 	response, err := ds.zabbixAPIRequest(ctx, dsInfo.GetUrl(), "apiinfo.version", map[string]interface{}{}, "")
 	if err != nil {
 		ds.logger.Debug("TestConnection", "error", err)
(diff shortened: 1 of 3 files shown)


pkg/zabbix_api.go
```diff
@@ -83,9 +84,9 @@ func (ds *ZabbixDatasource) ZabbixAPIQuery(ctx context.Context, tsdbReq *datasou
 		apiMethod := jsonQuery.Get("method").MustString()
 		apiParams := jsonQuery.Get("params").MustMap()
 
-		response, err := ds.ZabbixRequest(ctx, dsInfo, apiMethod, apiParams)
-		ds.queryCache.Set(HashString(tsdbReq.String()), response)
-		result = response
+		var err error
+		result, err = ds.ZabbixRequest(ctx, dsInfo, apiMethod, apiParams)
```
vignesh-reddy: I believe as part of this comment https://github.com/alexanderzobnin/grafana-zabbix/pull/815#discussion_r357008668 we removed ":=" which breaks the plugin. But the next commit added this back so went ahead fixed it.

pkg/zabbix_api_test.go
```diff
@@ -1 +1,232 @@
...
+}
+
+func TestZabbixRequestWithNoAuthToken(t *testing.T) {
+	var mockDataSource = mockZabbixDatasource{
+		ZabbixDatasource{
+			queryCache: NewCache(10*time.Minute, 10*time.Minute),
+			httpClient: NewTestClient(func(req *http.Request) *http.Response {
+				return &http.Response{
+					StatusCode: 200,
+					Body:       ioutil.NopCloser(bytes.NewBufferString(`{"result":"auth"}`)),
```
alecxvs: Is this supposed to test cases where auth is required? I would expect it to return a 400 error if so, and to somehow check that we make a call to get the auth token and try again.

alexanderzobnin:
Could you rebase it, please?

Issue #801, opened 2019-09-26
Backend: initial alerting query


alexanderzobnin:
In order to make alerting working, querying data source target should be reimplemented on the backend side. This makes sense only for time series data, so first part may be implementing `queryNumericData` method from the frontend.
For this purpose, a new query type should be created and used as a default. Main problem is implementing query parser which should fetch groups, hosts, apps, items and then parse regex filters to recognize items which should be queried.
https://github.com/alexanderzobnin/grafana-zabbix/blob/6e80b491205168a2744970bfd0403a76447e580b/pkg/datasource.go#L29-L34

alexanderzobnin:
@vignesh-reddy that's quite a lot of work. Try to take a look at `queryNumericData` method on the frontend:
https://github.com/alexanderzobnin/grafana-zabbix/blob/ca50d0db98638f079256417689a15fc3672d1f84/src/datasource-zabbix/datasource.js#L172
In order to get data from zabbix, data source should parse query first by fetching groups, hosts, apps and items.

PR #847, opened 2019-12-21
Backend: implement alerting query

vignesh-reddy:
Issue #801

Implemented alerting query issue to get data from Zabbix and parse the query by fetching groups, hosts, apps, and items.


pkg/models.go
```diff
@@ -9,3 +14,77 @@ type dbConnectionStatus struct {
...
+	ItemIDs     []string `json:"itemids,omitempty"`
+	GroupIDs    []string `json:"groupids,omitempty"`
+	HostIDs     []string `json:"hostids,omitempty"`
+	AppIDs      []string `json:"applicationids,omitempty"`
+
+	// Host Group GET
+	RealHosts bool `json:"real_hosts,omitempty"`
+
+	// History GET
+	History int `json:"history,omitempty,string"`
```
alexanderzobnin: This won't work for history type `0` since integer `0` will be treated as an empty value. Use `*int` instead.
```suggestion
	History *int `json:"history,string,omitempty"`
```

alexanderzobnin:
https://github.com/alexanderzobnin/grafana-zabbix/pull/847/files#diff-2cb344693fe3302e52c776fc8f6e3964R64
Datasource should be written in cache before returning:
```go
newDs := b.newZabbixDatasource()
b.datasourceCache.Set(dsInfoHash, newDs)
return b.newZabbixDatasource()
```

pkg/zabbix_api.go
```diff
@@ -275,3 +280,336 @@ func isNotAuthorized(message string) bool {
...
+			TimeFrom:  timeRange.GetFromEpochMs() / 1000,
+			TimeTill:  timeRange.GetToEpochMs() / 1000,
+		}
+
+		var response *simplejson.Json
+		var err error
+		if useTrend {
+			response, err = ds.ZabbixRequest(ctx, tsdbReq.GetDatasource(), "trend.get", params)
+		} else {
+			params.History = k
```
alexanderzobnin: History should be a pointer in order to prevent it to be omitted.
```suggestion
			params.History = &k
```

pkg/zabbix_api.go
```diff
@@ -275,3 +280,336 @@ func isNotAuthorized(message string) bool {
...
+	}
+
+	firstQuery := jsonQueries[0]
+
+	groupFilter := firstQuery.GetPath("group", "filter").MustString()
+	hostFilter := firstQuery.GetPath("host", "filter").MustString()
+	appFilter := firstQuery.GetPath("application", "filter").MustString()
+	itemFilter := firstQuery.GetPath("item", "filter").MustString()
+
+	items, err := ds.getItems(ctx, tsdbReq.GetDatasource(), groupFilter, hostFilter, appFilter, itemFilter, "num")
```
alexanderzobnin: `getItems()` return empty list in my case. Looks like something goes wrong there.

pkg/zabbix_api.go
```diff
@@ -275,3 +280,336 @@ func isNotAuthorized(message string) bool {
...
+	var items zabbix.Items
+	err = json.Unmarshal(itemsJSON, &items)
+	if err != nil {
+		return nil, err
+	}
+
+	filteredItems := zabbix.Items{}
+	for _, item := range items {
+		if item.Status == "0" {
+			matched, err := regexp.MatchString(itemFilter, item.Name)
```
alexanderzobnin: This filter doesn't work. There're 2 cases in filters:
1. Filter is a group/host/app/item name. So exact string comparison should be used.
1. Filter is a regex (wrapped in `/`, like `/CPU (.*) time/`). Since go regex should be compiled without first and last `/`, it should be removed before `MatchString()` execution.

See https://github.com/alexanderzobnin/grafana-zabbix/blob/master/src/datasource-zabbix/zabbix/zabbix.js#L417
```js
function findByFilter(list, filter) {
  if (utils.isRegex(filter)) {
    return filterByRegex(list, filter);
  } else {
    return findByName(list, filter);
  }
}
```

pkg/models.go
alecxvs: 9511524

pkg/zabbix_api.go
alecxvs: 9511524

alecxvs:
Caching new datasource instances: 74b97b6

pkg/zabbix_api.go
alecxvs: We're working on the Regex filter now. I pointed this out earlier but we set it aside to get a PR out before the holidays. The trickiest bit right now is handling flags, which a few are currently supported but the syntax for that is different in Go.

pkg/zabbix_api.go
alecxvs: We're able to get items with specific queries. Could this be related to the Regex filter not working properly?

pkg/zabbix_api.go
vignesh-reddy: I have updated the Regex filter the following way:
1) Check if the filter is a regex 
2) If it is a regex check for any flags, separate them and covert them to go regex format
3) If it is not regex do a regular string comparison

https://github.com/alexanderzobnin/grafana-zabbix/pull/847/commits/8bb98ae7a561048a307c4f13f214f4e39cd7d2e0

pkg/zabbix_api.go
vignesh-reddy: If any regex was used then it would never return any items, with the latest change that should be resolved.

pkg/zabbix_api.go
```diff
@@ -429,14 +440,20 @@ func (ds *ZabbixDatasource) getHosts(ctx context.Context, dsInfo *datasource.Dat
 	if err != nil {
 		return nil, err
 	}
+	isRegex, hostFilter := parseFilter(hostFilter)
+	re := regexp.MustCompile(hostFilter)
```
alecxvs: Don't use `MustCompile` if the filter isn't meant to be a regex. If it's not a valid pattern, it will panic.

Also perhaps we want to use something that returns an error when it fails to compile so we can surface that to the user as a user error.

vignesh-reddy: Updated, 04b1b2a

alecxvs: b6c2116
Cleaned it up a bit and added unit tests for parseFilter

pkg/zabbix_api.go
alecxvs: Also b6c2116
Added unit tests and added a check for supported flags.

PR #858, opened 2020-01-07
fix not acknowledged problem color with a message

### LINKED ISSUE #857: Severity color is changed to acknowleged even if there is only message without acknowlegement
From Zabbix 4.x there is a possibility to add a message to a problem without problem acknowlegement. These problems are shown with acknowleged color instead of severity color in Zabbix Problems view. Both layouts: table and list are affected.

A message assigned to a problem without problem acknowlegement should not change severity color to acknowledge color.

memfiz:
This PR closes #857.

As described in the issue #857 if a problem contains a message without the acknowledgment, then the severity color is shown as acknowleged in both views: table and list.

Zabbix [event object](https://www.zabbix.com/documentation/current/manual/api/reference/event/get) has property "acknowledged" which is 1 if the event has been acknowledged. I've used this property instead of "acknowledges" object, because acknowledges will contain the message attached to the event even if the event is not acknowleged.

### DIFF
--- a/src/panel-triggers/components/AlertList/AlertCard.tsx
+++ b/src/panel-triggers/components/AlertList/AlertCard.tsx
@@ -72,7 +72,7 @@ export default class AlertCard extends PureComponent<AlertCardProps, AlertCardSt
     let problemColor: string;
     if (problem.value === '0') {
       problemColor = panelOptions.okEventColor;
-    } else if (panelOptions.markAckEvents && problem.lastEvent.acknowledged == "1") {
+    } else if (panelOptions.markAckEvents && problem.lastEvent.acknowledged === "1") {
       problemColor = panelOptions.ackEventColor;
     } else {
       problemColor = severityDesc.color;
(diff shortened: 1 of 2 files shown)


src/panel-triggers/components/AlertList/AlertCard.tsx
```diff
@@ -72,7 +72,7 @@ export default class AlertCard extends PureComponent<AlertCardProps, AlertCardSt
     let problemColor: string;
     if (problem.value === '0') {
       problemColor = panelOptions.okEventColor;
-    } else if (panelOptions.markAckEvents && problem.acknowledges && problem.acknowledges.length) {
+    } else if (panelOptions.markAckEvents && problem.lastEvent.acknowledged == "1") {
```
alexanderzobnin: Should be strict `===`
```suggestion
    } else if (panelOptions.markAckEvents && problem.lastEvent.acknowledged === "1") {
```

PR #875, opened 2020-01-14
[Backend] Merge 'master' changes into the backend

alecxvs:
Merging master into the backend + some compilation fixes


package.json
```diff
@@ -25,21 +26,23 @@
...
-    "@grafana/toolkit": "^6.4.3",
-    "@grafana/ui": "^6.4.3",
+    "@babel/core": "^7.8.3",
+    "@babel/preset-env": "^7.8.3",
+    "@babel/preset-react": "^7.8.3",
+    "@emotion/core": "^10.0.27",
+    "@grafana/data": "canary",
+    "@grafana/runtime": "canary",
+    "@grafana/toolkit": "canary",
+    "@grafana/ui": "canary",
```
alecxvs: There's currently an issue that's failing unit tests if we don't mock @grafana/ui and @grafana/runtime: https://github.com/grafana/grafana/issues/20921

However there's also apparently an issue in master that's causing a different error entirely. Neither one works, but we're going to need 6.6 (or whatever release has the fix).

Basically Jest tests are broken because of Grafana packages, we can't fix it and need to wait for a good version of @grafana

package.json
```diff
@@ -25,21 +26,23 @@
...
     "@types/classnames": "^2.2.9",
     "@types/grafana": "github:CorpGlory/types-grafana",
     "@types/jest": "^23.1.1",
     "@types/jquery": "^3.3.0",
     "@types/lodash": "^4.14.104",
     "@types/moment": "^2.13.0",
-    "@types/react": "^16.4.6",
-    "@types/react-dom": "^16.0.11",
+    "@types/react": "^16.9.17",
+    "@types/react-table": "^6.8.6",
```
alecxvs: Current builds of @grafana/ui started using react-table 7. Without defining our own version of the types here, it was using types 7.x and throwing errors due to type changes in the next version.

package.json
alexanderzobnin: Aha, ok, that makes sense. I'll try to investigate. It seems, I remember some build error in grafana/ui related to Table interface.

PR #876, opened 2020-01-15
Added data source column for table layout

### LINKED ISSUE #700: Show Data Source name in Zabbix Problems Virtualization
I have multiple Zabbix Servers which everyone are monitoring the same hosts but with different IPs and from all around the world and some hostname are routed to multiple servers based on their geography, therefore all Zabbix Servers have the same configuration (hosts, host groups, triggers, etc.).

As we didn't want to redundant multiple zabbix nodes on main zabbix server we didn't use zabbix proxies, so it's not an option.

The issue is when I add multiple data sources and I see a problem I can't figure out which zabbix server is giving me this error, therefore I have to check manually again from zabbix servers' dashboard.

It would be so appreciated if there was an option which data source name could be added to dashboard like host name, technical name, host groups, etc.

memfiz:
This PR is just a suggestion thas closes #700

- Added dataSourceField panel option
- Added column "Data Source" for table layout
- As list layout already shows "Data Source" field it would be logical to show/hide data source based on the same new option. Added that.
- So far the option is made "true" by default, because it always was in list. However after an update the column will be visible to everyone that uses table layout. Other way is to use that new option by default as false and only for table layout, thus it will not change anything after a plugin update.

### DIFF
--- a/src/panel-triggers/triggers_panel_ctrl.js
+++ b/src/panel-triggers/triggers_panel_ctrl.js
@@ -62,7 +62,7 @@ export const PANEL_DEFAULTS = {
   statusIcon: false,
   severityField: true,
   ageField: false,
-  dataSourceField: true,
+  dataSourceField: false,
   descriptionField: true,
   descriptionAtNewLine: false,
   // Options
--- a/src/panel-triggers/triggers_panel_ctrl.js
+++ b/src/panel-triggers/triggers_panel_ctrl.js
@@ -62,6 +62,7 @@ export const PANEL_DEFAULTS = {
   statusIcon: false,
   severityField: true,
   ageField: false,
+  dataSourceField: true,
   descriptionField: true,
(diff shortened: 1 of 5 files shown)


src/panel-triggers/triggers_panel_ctrl.js
```diff
@@ -62,6 +62,7 @@ export const PANEL_DEFAULTS = {
   statusIcon: false,
   severityField: true,
   ageField: false,
+  dataSourceField: true,
```
alexanderzobnin: I don't think it makes sense to enable it by default, it helpfull for multi-datasource panel which is not enabled by default.

memfiz: I agree, I've made it false at the beginning, but then joining that option together with list layout made me change that to true...

Issue #891, opened 2020-02-13
ITEM.VALUE acts like ITEM.LASTVALUE in annotations


Akint:
**Describe the bug**
When {ITEM.VALUE} macros is used in a trigger name and annotations of this trigger occurs multiple times on a single graph, the macros is resolved like `{ITEM.LASTVALUE}` on every occurrence.
As you can see on the screenshot below, annotation text is always `too large: 5` for a trigger with condition `item.last() > 1` and title `too large: {ITEM.VALUE}`.
![old](https://user-images.githubusercontent.com/2721104/74419156-0d0ba680-4e7c-11ea-8f8c-90a2953cdc87.png)

**Expected behavior**
When possible, the historical (at-the-time-of-event) value should be seen in annotation text.
The screenshot below is a spoiler of how it works after my fix.
![expected](https://user-images.githubusercontent.com/2721104/74419200-1e54b300-4e7c-11ea-8c3e-0f0bfcb61839.png)

**Software versions**

| Grafana | Zabbix | Grafana-Zabbix Plugin |
| ------- | ------ | --------------------- |
| 6.6.1   | 4.0.17 | 3.10.5                |

P.S. May be a duplicate of #396, but the latter was never resolved.

alexanderzobnin:
Hi! Yes, it's better to use event name for annotations, like you did in your commit, good work!

PR #892, opened 2020-02-13
Fix ITEM.VALUE act like ITEM.LASTVALUE in annotations (#891)

Akint:
Before Zabbix 4.0 the described behavior was unpleasant but expected.
From [docs for {ITEM.VALUE}](https://www.zabbix.com/documentation/3.0/manual/appendix/macros/supported_by_location):
> Resolved to either:
> 1) the historical (at-the-time-of-event) value of the Nth item in the trigger expression, if used in the context of trigger status change, for example, when displaying events or sending notifications.
> 2) the latest value of the Nth item in the trigger expression, if used without the context of trigger status change, for example, when displaying a list of triggers in a pop-up selection window. In this case works the same as {ITEM.LASTVALUE}

But this changes with Zabbix 4.0+. The macros resolution works the same, but [Event object](https://www.zabbix.com/documentation/4.0/manual/api/reference/event/object) now has a new property `name` that contains event name resolved in the correct way (using at-the-time-of-event value).

### DIFF
--- a/src/datasource-zabbix/datasource.js
+++ b/src/datasource-zabbix/datasource.js
@@ -536,12 +536,13 @@ export class ZabbixDatasource {
             let title = Number(event.value) ? 'Problem' : 'OK';
 
             let formatted_acknowledges = utils.formatAcknowledges(event.acknowledges);
+            let event_name = this.zabbixVersion < 4 ? indexedTriggers[event.objectid].description : event.name;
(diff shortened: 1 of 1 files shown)


alexanderzobnin:
Fixed in 8ac5f42e73808468a75f07d4f1250a9a435f85a6

Issue #1131, opened 2020-12-30
Grafana / Maintenance Zabbix


luedthsa:
Under normal conditions when referring to maintenance, they are created in zabbix including at least one host or a group of hosts and in the same configuration are set the triggers TAGs that should be omitted in the graphana dashboard, however both in graphana and in zabbix, the host or host group assigned at the beginning of the configuration receives a maintenance tag that is represented by orange pliers, in zabbix when the checkbox for displaying suppressed incidents is not checked, the host alerts that do not have the maintenance tag appear normally, but the host is displayed with the tag of the pliers, when checked it shows all alerts from the host including the alerts of tags under maintenance, however the graphana does not have this same interpretation, and it omits everything that has the pliers or displays everything that has the pliers. Is there a workaround for this?

![image](https://user-images.githubusercontent.com/62388006/103368976-bc0e9c00-4aa7-11eb-9a86-3ea10ad2cbb4.png)
![image](https://user-images.githubusercontent.com/62388006/103368988-c466d700-4aa7-11eb-9f07-c2f2fc3c2826.png)

luedthsa:
![WhatsApp Image 2020-12-29 at 18 36 32](https://user-images.githubusercontent.com/62388006/103359514-d429ef80-4a96-11eb-8e22-32e172620d2f.jpeg)
![WhatsApp Image 2020-12-29 at 18 35 54](https://user-images.githubusercontent.com/62388006/103359519-d55b1c80-4a96-11eb-9f35-61f08890f77e.jpeg)

luedthsa:
![WhatsApp Image 2020-12-29 at 18 36 32](https://user-images.githubusercontent.com/62388006/103359822-5e725380-4a97-11eb-8dbf-dee80a748c80.jpeg)
![WhatsApp Image 2020-12-29 at 18 35 54](https://user-images.githubusercontent.com/62388006/103359824-5fa38080-4a97-11eb-8179-f5e95ecb341b.jpeg)

luedthsa:
![maintenance tag](https://user-images.githubusercontent.com/62388006/103368390-2b838c00-4aa6-11eb-9733-1d92e2d2ff72.png)
![maintenance tag2](https://user-images.githubusercontent.com/62388006/103368393-2c1c2280-4aa6-11eb-9f02-838825469801.png)

hudnall-s:
[...]

Issue #830, opened 2019-11-08
Show non-suppressed problems from maintenance hosts even when "Show Maintenance Hosts" is not selected


TomasManzano:
The ability to set maintenance periods to triggers matching tags instead of the whole host is a great step forward for Zabbix.

When a non-suppressed trigger of a maintenance host fires up the Problems tab in Zabbix shows this problem as the logic has been updated and now allows to hide "suppressed problems" instead of "maintenance hosts"

In grafana-zabbix if "Show hosts in maintenance" is not selected, unsuppressed problems from maintenance hosts won't appear on the dashboard. A workaround is to select "Show hosts in maintenance" but that fills the dashboard with a bunch of no use problems that make the approach unfeasible.  IE if you have stopped a complete environment for some reasons and you have also got some services in scheduled maintenance.  That situation happens a lot over here.

cristianoness:
@alexanderzobnin We have the same problem here, some solution in mind?

hudnall-s:
for anyone else looking for this, I rebuilt the plugin to use the maintenance mode on a problem level instead of a host level.

current as of 2/9/2022 can be found here https://github.com/hudnall-s/Grafana-Zabbix-Plugin

PR #1393, opened 2022-02-12
change maintenance windows logic to work with tagging (problem level) solves #830, #1131

Sam-H101:
Problem:
Currently for companies that use Tagging with Maintenance Windows, the current logic just does a host blanket blackout. Changed the logic to look on the problem level instead of the host level. This will use zabbix's logic to determine if a particular problem is truly under Maintenance or not.

Solves:
[Issue 830](https://github.com/alexanderzobnin/grafana-zabbix/issues/830)
[Issue 1131](https://github.com/alexanderzobnin/grafana-zabbix/issues/1131)


Sam-H101:
@alexanderzobnin any thoughts on this? Still retains old "host only" maintenance but also is smarter so if there is tagging, it will only throw those in maintenance.

alexanderzobnin:
Hi! I couldn't merge this due to merge conflicts, but I copied this and added you as a co-author. Thank you for the contribution!

Issue #1465, opened 2022-06-17
Application error. Script is not allowed in manual host action: scope:4


miyagidaniel:
Is possible execute script scope "manal event action" on zabbix problems panel ?

leogiraldimg:
Solves: [Issue 1465](https://github.com/alexanderzobnin/grafana-zabbix/issues/1465)

rodrigocastrillon:
+1

I can see the scrips listed in the alerts panel, but the it seems that only "manual host action" is supported, not "manual event action". I use tags in the event to help the script to receive specific information on that is being alerted, so this would be a very nice feature to have.

AhmedALSGit:
It seems that this issue has not been patched on the newer version of the plugin (4.4.0), could you please find a solution to use the action on a "manual event action". It would help us a lot !
![image](https://github.com/grafana/grafana-zabbix/assets/56633004/b6f801ae-e170-417e-885a-4a8ddb002bae)

AhmedALSGit:
Any update on this case or workarounds on new releases ?
Thanks in advance

ochmite:
Hi everyone,
I’d like to request the addition of script execution in the "manual event action" of Zabbix Problems panel. This functionality is vital for managing alerts effectively and has been previously discussed (Issue #1465), but it remains unresolved in recent plugin versions :( 
Please consider prioritizing this feature in future updates, it would greatly enhance the plugin's usability.
Thanks for your efforts and attention to this

jbew81:
Hi Grafana-Zabbix Team,

As everyone in this thread, i am extremely frustrated 😡😡😡 with this persistent issue. This feature is vital for my operations, and its absence is causing significant disruptions. 
It is unacceptable that this problem remains unresolved despite being raised multiple times. Additionally, this dysfunction is preventing me from upgrading to the latest release of the plugin 👎 . 
Is it possible that this issue (that seems to be simple to fix) be prioritized and fixed immediately ?

majax999:
Hello,

I'm also experiencing the issue, this problem is preventing me from upgrading to the latest plugin release. 
Please fix this as soon as possible.

AhmedALSGit:
Hello,

I temporarily resolved this issue by patching the plugin with the modifications below :
- Get the scope of the script (manual event or manual host)
- Execute the script and add event ID or host ID based on the scope.

The change was implemented in version 4.6.1, which is the most recent version as of now.

[...]

PR #1466, opened 2022-06-21
feat: enable executing manual event scripts

leogiraldimg:
Solves: [Issue 1465](https://github.com/alexanderzobnin/grafana-zabbix/issues/1465)


leogiraldimg:
We've tested the new feature with our development environment:

* The action of executing a manual event script before the implementation: image
* The action of executing a manual event script after the implementation: image

PR #1472, opened 2022-07-07
Switch to regexp2 for parseFilter()

### LINKED ISSUE #1264: Perl regex syntax no longer accepted in any filter (plugin 4.2.1, Grafana 8.1.1)
**Describe the bug**
After update to 4.2.1 plugin (and grafana 8.1.1) any panel with an item/application/host/group filter that uses Perl regex syntax will show no-data. Using simpler regex syntax shows data as normal. Tried to switch from old "Graph" viz to the new "Time series" without any change in behaviour. 

**Expected behavior**
Time-series graph shown also for Perl regex syntax filters as the same syntax worked previously (grafana 7.x/8.0 and plugin 4.1.5). Edit: it also works with grafana 8.1.2+plugin 4.1.5 (but that has other issues)

**Network data**

{"queries":[{"application":{"filter":"CPU"},"countTriggers":true,"functions":[],"group":{"filter":"/^Linux servers$/"},"host":{"filter":"/^mgc01$/"},"item":{"filter":"/^CPU (?!idle|usage|ready)/"},"minSeverity":3,"options":{"disableDataAlignment":false,"showDisabledItems":false,"skipEmptyValues":false,"useZabbixValueMapping":false},"proxy":{"filter":""},"queryType":"0","refId":"A","resultFormat":"time_series","table":{"skipEmptyValues":false},"tags":{"filter":""},"trigger":{"filter":""},"triggers":{"acknowledged":2,"count":true,"minSeverity":3},"itemTag":{"filter":"Application: CPU"},"datasource":"Zabbix","datasourceId":1,"intervalMs":10000,"maxDataPoints":931}],"range":{"from":"2021-08-12T05:34:20.111Z","to":"2021-08-12T08:34:20.111Z","raw":{"from":"now-3h","to":"now"}},"from":"1628746460111","to":"1628757260111"}

{"results":{"A":{"error":"error parsing regexp: invalid or unsupported Perl syntax: `(?!`","frames":[]}}}

**Software versions**

| 8.1.1   | 5.2.7  | 4.2.1               |

thetuxkeeper:
I changed the returned regexp of `parseFilter()` to regexp2. I kept all other uses `regexp` since there's no need for a switch and `regexp2` seems to be unnecessary complex to use for something like `FindStringSubmatch()`.
I haven't run any tests on a real grafana instance, I'll try that later. But I extended the tests and they passed.

If there's anything I could/should change, just let me know! :)

This PR should fix #1264

### DIFF
--- a/pkg/zabbix/methods.go
+++ b/pkg/zabbix/methods.go
@@ -148,7 +148,7 @@ func filterItemsByTag(items []*Item, filter string) ([]*Item, error) {
 			}
 			for _, t := range tags {
 				if re != nil {
-					if re.MatchString(t) {
+					if match, err := re.MatchString(t); match && err != nil {
 						filteredItems = append(filteredItems, i)
 						break
 					}
@@ -173,7 +173,7 @@ func filterItemsByQuery(items []*Item, filter string) ([]*Item, error) {
 	for _, i := range items {
 		name := i.Name
 		if re != nil {
-			if re.MatchString(name) {
+			if match, err := re.MatchString(name); match && err != nil {
 				filteredItems = append(filteredItems, i)
 			}
 		} else if name == filter {
@@ -212,7 +212,7 @@ func filterAppsByQuery(items []Application, filter string) ([]Application, error
 	for _, i := range items {
 		name := i.Name
 		if re != nil {
-			if re.MatchString(name) {
+			if match, err := re.MatchString(name); match && err != nil {
 				filteredItems = append(filteredItems, i)
 			}
 		} else if name == filter {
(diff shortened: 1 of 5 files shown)


pkg/zabbix/methods.go
```diff
@@ -173,7 +173,7 @@ func filterItemsByQuery(items []*Item, filter string) ([]*Item, error) {
 	for _, i := range items {
 		name := i.Name
 		if re != nil {
-			if re.MatchString(name) {
+			if match, err := re.MatchString(name); match && err != nil {
```
alexanderzobnin, line 176: That should be  like
```suggestion
			if match, err := re.MatchString(name); match && err == nil {
```

PR #1493, opened 2022-08-29
Problems count mode

alexanderzobnin:
Problems count mode.


src/datasource-zabbix/partials/query.editor.html
```diff
@@ -11,7 +11,7 @@
                 </select>
             </div>
         </div>
-        <div class="gf-form" ng-show="ctrl.target.queryType == editorMode.TEXT">
+        <div class="gf-form" ng-show="ctrl.target.queryType == editorMode.TEXT | ctrl.target.queryType == editorMode.MACROS">
```
alexanderzobnin, line 14:
```suggestion
        <div class="gf-form" ng-show="ctrl.target.queryType == editorMode.TEXT || ctrl.target.queryType == editorMode.MACROS">
```

Issue #1386, opened 2022-02-08
Error "json: invalid use of ,string struct tag, trying to unmarshal unquoted value into int64"


Ethan0451:
Grafana | Zabbix | Grafana-Zabbix Plugin
-- | -- | --
8.3.1 | 5.4.9 | 4.2.4

A few days ago , I updated the zabbix from 5.2.2 to 5.4.9
At the same time , I reinstall the grafana and grafana-zabbix plugin , but when I create the panel , I got this error
By the way , my zabbix use Elasticsearch to store history data

Thanks for helping

[grafana.log](https://github.com/alexanderzobnin/grafana-zabbix/files/8020338/grafana.log)
[request_details.txt](https://github.com/alexanderzobnin/grafana-zabbix/files/8020379/request_details.txt)

nikossss:
solved the problem? please help me

nikossss:
yes, i m use elastic. thank you, i replace a few line, how to compiled?

nikossss:
please tell me how to compiled?

Ethan0451:
> please tell me how to compiled?

https://alexanderzobnin.github.io/grafana-zabbix/installation/run_from_master/

nikossss:
ansible@ansible:~/grafana-zabbix-4.2.9$ make dist
yarn build
00h00m00s 0/0: : ERROR: [Errno 2] No such file or directory: 'build'
make: *** [Makefile:42: dist-frontend] Error 1
ansible@ansible:~/grafana-zabbix-4.2.9$ sudo make dist
yarn build
00h00m00s 0/0: : ERROR: [Errno 2] No such file or directory: 'build'
make: *** [Makefile:42: dist-frontend] Error 1
don't work

nikossss:
don,t work(

    Entrypoint undefined = extract-text-webpack-plugin-output-filename
    [../node_modules/css-loader/dist/cjs.js!../node_modules/sass-loader/dist/cjs.js!./sass/grafana-zabbix.dark.scss] 41.1 KiB {0} [built]
        + 2 hidden modules
Child extract-text-webpack-plugin ../node_modules/extract-text-webpack-plugin/dist ../node_modules/css-loader/dist/cjs.js!../node_modules/sass-loader/dist/cjs.js!sass/grafana-zabbix.light.scss:
    Entrypoint undefined = extract-text-webpack-plugin-output-filename
[...]

PR #1494, opened 2022-08-30
Update models.go

flexboy:
Error "json: invalid use of ,string struct tag, trying to unmarshal unquoted value into int64"
#1386

### DIFF
--- a/pkg/zabbix/models.go
+++ b/pkg/zabbix/models.go
@@ -67,8 +67,8 @@ type ItemTag struct {
 type Trend []TrendPoint
 
 type TrendPoint struct {
-	ItemID   string `json:"itemid,omitempty"`
-	Clock    int64  `json:"clock,omitempty,string"`
+	ItemID   int64 `json:"itemid,omitempty"`
+	Clock    int64  `json:"clock,omitempty"`
 	Num      string `json:"num,omitempty"`
 	ValueMin string `json:"value_min,omitempty"`
 	ValueAvg string `json:"value_avg,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/zabbix/models.go
```diff
@@ -67,8 +67,8 @@ type ItemTag struct {
 type Trend []TrendPoint
 
 type TrendPoint struct {
-	ItemID   string `json:"itemid,omitempty"`
-	Clock    int64  `json:"clock,omitempty,string"`
+	ItemID   int64 `json:"itemid,omitempty"`
```
alexanderzobnin, line 70:
```
pkg/datasource/response_handler.go:24:24: cannot use point.ItemID (type int64) as type string in map index
pkg/datasource/response_handler.go:25:15: cannot use point.ItemID (type int64) as type string in map index
pkg/datasource/response_handler.go:26:13: cannot use point.ItemID (type int64) as type string in map index
pkg/datasource/response_handler.go:28:27: cannot use point.ItemID (type int64) as type string in map index
```
But this is a good idea, I'll take a look closer.

flexboy: 24,26c24,26
< 		pointItem := itemsMap[point.ItemID]
< 		if seriesMap[point.ItemID] == nil {
< 			seriesMap[point.ItemID] = timeseries.NewTimeSeriesData()
---
> 		pointItem := itemsMap[strconv.FormatInt(point.ItemID,10)]
> 		if seriesMap[strconv.FormatInt(point.ItemID,10)] == nil {
> 			seriesMap[strconv.FormatInt(point.ItemID,10)] = timeseries.NewTimeSeriesData()
28c28
< 		pointSeries := seriesMap[point.ItemID]
---
>

alexanderzobnin:
Are you able to replicate the original bug?

Elufimov:
I am not an author but I have an original bug with grafana 9.2.0 zabbix 6.2.0 and elasticsearch 7.13.2.

Issue #1315, opened 2021-09-24
Performance Improvement ItemTag


erik-de-neve:
After migrating to Zabbix 5.4 we see a huge performance impact on Grafana.
With Zabbix 5.4 (where applications are replaced by item tags) our Grafana dashboards are loading much slower (some of them even time out).

Before Zabbix 5.4 you filtered on the application id 'appids' if an application was filled in.
But since Zabbix 5.4 applications no longer exist, so the value will always be 'nil'.

Looking into the code in pkg/zabbix/methods.go I see this:
```go
var allItems []*Item
if len(appids) > 0 {
  allItems, err = ds.GetAllItems(ctx, nil, appids, itemType)
} else if len(hostids) > 0 {
  allItems, err = ds.GetAllItems(ctx, hostids, nil, itemType)
}

if isZabbix54orHigher && itemTagFilter != "" {
  allItems, err = filterItemsByTag(allItems, itemTagFilter)
  if err != nil {
    return nil, err
  }
}
```

If I understand it well, you are querying all items without an itemTag filter and then you do the filtering on the Grafana side (func filterItemsByTag).
Before 5.4 the application filter was taking into account in this api call.

So now when we query on a lot of hosts and specify a specific item tag, it will still query all items on all of these hosts regardless of the item tag.
It is only afterwards that in the plugin itself you will filter on the item tag.
For us this makes a huge difference since sometimes we have 200 hosts with 5000 items each. That's a lot of items and a lot of data to query.

I think this is not really scalable as it returns too much items from Zabbix. 
Better would be to include the ItemTag filter directly in the Zabbix-api call instead of doing the filtering on the Grafana side after having returned a bunch of items.

**Software versions**

| Grafana | Zabbix | Grafana-Zabbix Plugin |
| ------- | ------ | --------------------- |
| 8.1.5   | 5.4.4  | 4.2.3                 |

alexanderzobnin:
Yes, you're right. Unfortunately, Zabbix does not provide a separate API method for pulling list of tags, so I have to get it from items. But I definitely can improve performance by querying only items with specific tags.

erik-de-neve:
Hello,

Do you have an idea when this could be implemented?
For us grafana is unusable for the moment, because we have many items on 1 host and the performance is really killing us due to the tags and how it's implemented.

Thanks!
Erik

alexanderzobnin:
Fixed in c1b0d963cd7375b69e050b91f3989be87cecc16b

PR #1526, opened 2022-11-17
improve tags performance

erik-de-neve:
closes https://github.com/alexanderzobnin/grafana-zabbix/issues/1315.


alexanderzobnin:
Merged manually. Added @erik-de-neve as a co-author.

PR #1686, opened 2023-09-11
Solve Grafana's inability to obtain data when using Elasticsearch as historical data storage

flexboy:
Solve Grafana's inability to obtain data when using Elasticsearch as historical data storage,


Elufimov:
It would be nice to have this, right now I forced to apply manually this path in every upgrade.

Elufimov:
This is exact problem as where here https://github.com/grafana/grafana-zabbix/pull/1514  It would be nice is this pr be merged/

PR #1688, opened 2023-09-13
track executed queries and panel clicks

gwdawson:
track executed queries so we can view some stats around the datasource
this includes but not limited to:
- how many queries are sent
- how often each query type is used
- etc.

Fixes [#72899](https://github.com/grafana/grafana/issues/72899)

### DIFF
--- a/src/datasource/tracking.ts
+++ b/src/datasource/tracking.ts
@@ -1,6 +1,15 @@
 import { DataQueryRequest } from '@grafana/data';
 import { ZabbixMetricsQuery } from './types';
 import { reportInteraction } from '@grafana/runtime';
+import {
+  MODE_ITEMID,
+  MODE_ITSERVICE,
+  MODE_MACROS,
+  MODE_METRICS,
+  MODE_PROBLEMS,
+  MODE_TEXT,
+  MODE_TRIGGERS,
+} from './constants';
 
 export const trackRequest = (request: DataQueryRequest<ZabbixMetricsQuery>): void => {
   request.targets.forEach((target) => {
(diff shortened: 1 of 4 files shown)


src/datasource/tracking.ts
```diff
@@ -0,0 +1,65 @@
+import { DataQueryRequest } from '@grafana/data';
+import { ZabbixMetricsQuery } from './types';
+import { reportInteraction } from '@grafana/runtime';
+
+export const trackRequest = (request: DataQueryRequest<ZabbixMetricsQuery>): void => {
```
gabor, line 5: phew, this is a lot of code here 😄 .

i understand the reason.

still, i'd recommend one of two approaches:
1. only track some minimal info, like only the query-type
2. refactor this code:
    - here we can have a function that just builds the "report-interaction-object" from the request-object
       - we can write a bunch of unit-tests for it to be sure it's not crashing
    - and then call it from `datasource.ts` with something like `reportInteraction('something_something', extractProps(request))`


random note: when the zabbix code handles the `.queryType`, it seems to use these values to compare with: https://github.com/grafana/grafana-zabbix/blob/f7d84d1b1ff966c386627fa666cf884d5465ac71/src/datasource/constants.ts#L10-L16 (for example see here: https://github.com/grafana/grafana-zabbix/blob/f7d84d1b1ff966c386627fa666cf884d5465ac71/src/datasource/components/QueryEditor.tsx#L118 ), maybe we could use those values too?

gwdawson: i think we should just go with option 1 for now then and if we decide in future we would like more info we could implement 2.

gabor (approved):
LGTM 👍

PR #1692, opened 2023-09-24
Chore: Dependency clean up

zoltanbedi:
In this PR:
- Removed some of the old CI stuff and scripts related to them
- Updated create-plugin which made some changes in the .config folder
- Removed the e2e dependencies and removed one test that was not really testing anything
- Removed a couple of unused dependencies and bumped most of them to remove vulnerabilities
- Converted zabbix.test.js to typescript
- Removed some code in the jest-setup.js that were not needed anymore


src/test-setup/jest-setup.js
```diff
@@ -1,102 +1,123 @@
...
-jest.mock('grafana/app/core/core_module', () => {
-  return {
-    directive: function() {},
-  };
-}, {virtual: true});
-
-jest.mock('grafana/app/core/core', () => ({
-  contextSrv: {},
-}), {virtual: true});
+jest.mock(
```
zoltanbedi, line 3: This is changed because of prettier format

package.json
```diff
@@ -1,137 +1,101 @@
...
-  },
   "resolutions": {
-    "js-yaml": "^3.13.1",
-    "lodash": "~4.17.13",
-    "set-value": "^2.0.1",
-    "mixin-deep": "^1.3.2",
-    "minimatch": "^3.0.2",
-    "fstream": "^1.0.12"
-  }
+    "jackspeak": "2.1.1"
```
zoltanbedi, line 98: Had to pin this dependency because of an issue in yarn 1. See https://github.com/storybookjs/storybook/issues/22431#issuecomment-1630086092 for more info.

devenv/zabbix62/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: "3"
 services:
   # Grafana
   grafana:
-    image: grafana/grafana:8.4.7
+    image: grafana/grafana:10.1.2
```
gabor, line 6: there are multiple devenv-docker-compose files with `grafana` and `docker` sections.. i'd recommend that we either adjust all of the files, or we do not adjust any of them (and do the adjustments in a sepearate PR). WDYT?

devenv/zabbix62/docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
       POSTGRES_DB: zabbix
 
   database:
-    image: postgres
+    image: postgres:15
```
gabor, line 51: there are multiple devenv-docker-compose files with this postgres-entry.. do we want to fix all of them here?

gabor (approved):
very nice work. LGTM 👍

devenv/zabbix62/docker-compose.yml
zoltanbedi: yeah I think that makes sense.

zoltanbedi: FYI I had to bump this because postgres translates to the latest which is 16 and zabbix server shuts down with an error:
`Unable to start Zabbix server due to unsupported PostgreSQL database version (16.0). Must not be higher than (15.x)`

zoltanbedi:
I wonder if @alexanderzobnin is ok with these changes?

PR #1705, opened 2023-10-09
Bump 4.4.2

zoltanbedi:


gabor (approved):
LGTM 👍 
(good catch with the urls in the `package.json`)

Issue #1744, opened 2023-11-15
Cache doesn't hit for item.get api with tags


iqbalaydrus:
**Describe the bug**
Configured with 1h cache ttl, the consecutive queries are still hitting item.get zabbix api and not taking it from the cache.

**Expected behavior**
Same query should be using cache instead of api calls.

**Software versions**

| Grafana | Zabbix | Grafana-Zabbix Plugin |
| ------- | ------ | --------------------- |
| 9.4.17   | 6.4.8  | 4.4.3                 |

iqbalaydrus:
Closing this since PR got merged

PR #1745, opened 2023-11-15
item.get cache won't be hit if tags order are different

iqbalaydrus:
Solves #1744


iqbalaydrus:
Ok, I added unit test for this PR

zoltanbedi (approved):
LGTM thank you.

PR #1748, opened 2023-11-16
Updated backend plugin SDK

yesoreyeram:
Updating the backend SDK which contains otel dependency update.


gabor (approved):
LGTM 👍

(i think we'll have to update the CHANGELOG when we make the release, there may be a couple changes we do not have included there yet)

PR #1750, opened 2023-11-17
fix reset bug in config page

gwdawson:
after the configuration page updates the `Reset` button for the zabbix password stopped working.
this fixes that so the user can update the configured password.


gabor (approved):
LGTM 👍

PR #1770, opened 2024-01-08
feat: add generated jsonnet library

Duologic:
Fixes https://github.com/grafana/grafonnet/issues/149

In the Grafonnet community some folks asked for support on the Zabbix plugin. I'm taking the liberty to propose adding this to the repo but feel free to point me in another direction.

This PR generates a library from the Typescript type `ZabbixMetricsQuery`, I'm boldly assuming that this is the type representing the JSON that is visible in the Dashboard JSON. Note that anything in `jsonnet/zabbix-query-libsonnet/` is generated by the Makefile targets.

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -28,7 +28,11 @@
     "docs/**/*.yaml",
     "dashboards/**",
     "src/**",
-    "pkg/**"
+    "pkg/**",
+    "jsonnet/generator/jsonnetfile.json",
+    "jsonnet/generator/jsonnetfile.lock.json",
+    "jsonnet/generator/schema.json",
+    "jsonnet/zabbix-query-libsonnet/docs/**"
   ],
   "ignoreRegExpList": [
     "import\\s*\\((.|[\r\n])*?\\)",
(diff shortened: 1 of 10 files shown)


Duologic:
That's a good question. This initial PR simply syncs the schema with the types on the same branch, I'd suggest we match the schema sync that way along with the versions so it aligns with the release cadence.

For core plugins, this is currently happening through grafana/grok (using the CUE schemas) but we're looking in getting such JSON Schemas directly in the grafana/grafana repo from Typescript.

For other plugins, they could follow a similar path as this plugin, I'd say that this PR could serve as a pilot to generate jsonnet libraries near the plugin itself. The code in the `generator/main.libsonnet` is specific to this plugin but the pattern should be repeatable in other plugins.

At this point I'd suggest doing this for other plugins on a per-request basis as we don't want to introduce unnecessary work and code to maintain.

Duologic:
Would y'all be less opposed to only having to generate and maintain the JSON schema? Then we can move the library itself into Grafonnet while keeping up with the releases of this plugin.

Duologic:
The actual generator is not in here, that's the CRDsonnet package, the generator directory here holds the specific code for generating this library.

Duologic:
I've put this into a separate repo: https://github.com/grafana/grafana-zabbix-libsonnet/

PR #1800, opened 2024-03-11
config updates

yesoreyeram:
- [x] Updated author names in plugin.json files for sub components of the app plugin ( datasource and problem's panel's plugins.json )
- [x] Updated `.config` folder using create-plugin update
- [x] Updated docker compose file to use consistent version
- [x] Split the `types.ts` file into `types/query.ts` and `types/config.ts`


zoltanbedi (changes requested):

zoltanbedi (approved):
lgtm

PR #1822, opened 2024-04-22
fixes for scenes framework

yesoreyeram:
* Fixes https://github.com/grafana/grafana-zabbix/issues/1819
* Also update backend plugin SDK

### DIFF
--- a/src/datasource/zabbix/proxy/cachingProxy.ts
+++ b/src/datasource/zabbix/proxy/cachingProxy.ts
@@ -104,20 +104,13 @@ function cacheRequest(func, funcName, funcScope, self) {
 function getRequestHash(args) {
   try {
     if (typeof args === 'object') {
-      args = Array.from(args || {}).map((arg: any) => {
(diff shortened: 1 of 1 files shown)


src/datasource/zabbix/proxy/cachingProxy.ts
```diff
@@ -100,9 +100,25 @@ function cacheRequest(func, funcName, funcScope, self) {
...
+          return { ...rest, scopedVars: scopedVarsWithoutScenes };
+        }
+        return arg;
+      });
+    }
+    const argsJson = JSON.stringify(args);
+    return getHash(argsJson);
+  } catch (ex) {
+    console.error(`failing to get the request hash for caching. type of args ${typeof args}`);
+    return 0;
```
gabor, line 120: i wonder if it would be safer to return `Math.random()` or something similar here, because `0` will cause every such request to share the cache-key...  WDYT?

gabor (approved):
LGTM 👍

PR #1894, opened 2024-10-15
Bug: Fix `AlertList` styling

ashharrison90:
- as part of https://github.com/grafana/grafana/pull/90224, i removed these SCSS styles from core grafana
  - unfortunately the dashboard i was using to see which plugins were using which styles never flagged this plugin as using them 😭
  - still need to dig into how and why these usages weren't shown
- copy these styles into the plugin where they can be controlled and adjusted at your leisure 😄 

thanks to @PedroRibas13 @mplm17 for raising this 🙇 

Fixes https://github.com/grafana/grafana-zabbix/issues/1868

### DIFF
--- a/src/styles/_panel-triggers.scss
+++ b/src/styles/_panel-triggers.scss
@@ -1,3 +1,41 @@
+.alert-rule-list {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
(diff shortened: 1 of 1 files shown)


zoltanbedi (approved):
Thanks for fixing this for us @ashharrison90 🫶

Issue #1616, opened 2023-04-07
Authentication broken for ancient Zabbix since support added for Zabbix 6.4


colinleroy:
Zabbix changed their user.login API endpoint to use "username" instead of "user" in 6.4rc, and you added support for it.
You added a isDeprecatedUserParamError check in order to be retro-compatible, and you look for 'unexpected parameter "user' in the error string. So far so good.

But ancient Zabbix installs are now broken, because when getting a user.login request with "username" instead of "user", they throw a generic error: 

```
$ cat login.json; curl  --header 'Content-Type: application/json-rpc' -k https://zabbix-3.0-instance/zabbix/api_jsonrpc.php -d @login.json 
{
  "jsonrpc": "2.0",
  "method": "user.login",
  "params": {
    "username": "api_grafana",
    "password": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
  },
  "id": 1
}

{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params.","data":"Login name or password is incorrect."},"id":1}
```

Sadly, I have no idea at what point Zabbix changed from that generic error to the one where they check the parameter names.

Possibly it would be cleaner to check the apiinfo.version endpoint before trying to user.login, and set the parameter name accordingly. Unfortunately, I suck at go and don't think I can provide a good pull request.

Hope this helps!

theguimaraes:
I have the same issue here.

Grafana 9.4.3 and i can't connect to one of my old Zabbix 3.0

Plugin version 4.3.0 and 4.3.1, only connect to my new Zabbix 6.0, but can't connect to the 3.0

Plugin version 4.2.10 and lower, i can connect to my old Zabbix 3.0, but can't connect to the new 6.0. I receive a Error: Invalid params. Invalid parameter "/": unexpected parameter "user".

renatotoniolo:
I'm having the same problem, my Zabbix 3.0 in which I'm migrating to a new Zabbix 6.4 in parallel, I'm using two datasources to keep the alerts in a single dashboard and I'm having the same problem.

alexanderzobnin:
Fixed in 4.3 #1544

colinleroy:
> Fixed in 4.3 #1544

No, this is a different issue. The 4.3 makes the plugin work on Zabbix 6.4, but breaks the plugin with Zabbix 3.0

PR #1923, opened 2024-11-26
Simplifies authentication

wiltonsr:
- Remove the `Authenticate` and `isDeprecatedUserParamError` methods
- Using the Zabbix API version to determine the use of `Login` or `LoginDeprecated`
- This will prevent errors login in old Zabbix 3 as reported in #1616

### DIFF
--- a/pkg/zabbix/zabbix.go
+++ b/pkg/zabbix/zabbix.go
@@ -141,11 +141,20 @@ func (zabbix *Zabbix) Authenticate(ctx context.Context) error {
 		zabbixPassword = jsonData.Get("password").MustString()
 	}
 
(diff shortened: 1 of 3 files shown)


ivanahuckova (commented):
@wiltonsr there has been some changes related to support of zabbix v7. The `zabbix.api.Authenticate` now accepts version, so you could just use your code in that method.
 
```
if zabbix.version >= 54 {
		auth, err = zabbix.api.Login(ctx, zabbixLogin, zabbixPassword)
	} else {
		auth, err = zabbix.api.LoginDeprecated(ctx, zabbixLogin, zabbixPassword)
	}
```

PR #1960, opened 2025-02-03
Fix problems panel broken when selecting triggers

yesoreyeram:
This PR fixes an issue where the problems panel breaks when you change to query type to `Triggers`.
 

How to reproduce:

1. Go to `Zabbix data source features` dashboard
2. Open Problems panel
3. Change`Query type` to `Triggers`
4. Select `Problems` from `Count by`
5. Change the `Group` to `Backend`
6. The problems panel should not error


gwdawson (approved):
LGTM

PR #1967, opened 2025-02-12
Use `go install` to install golint

samjewell:
This is the recommended approach since Go 1.17

I was getting an error when running `make install`:
```
GO111MODULE=off go get -u golang.org/x/lint/golint
go: modules disabled by GO111MODULE=off; see 'go help modules'
```


ivanahuckova (approved):
Thank you!

PR #2003, opened 2025-03-28
add option to show data source name

ovdgInitworks:
https://github.com/grafana/grafana-zabbix/issues/1766

### DIFF
--- a/devenv/dashboards/zabbix_datasource_features.json
+++ b/devenv/dashboards/zabbix_datasource_features.json
@@ -37,6 +37,7 @@
         "uid": "PCAF1C06DCF802728"
       },
       "descriptionAtNewLine": false,
+      "showDatasourceName": false,
       "descriptionField": true,
       "fontSize": "100%",
       "gridPos": {
@@ -64,6 +65,7 @@
         "allowDangerousHTML": false,
         "customLastChangeFormat": false,
         "descriptionAtNewLine": false,
+        "showDatasourceName": false,
         "descriptionField": true,
         "fontSize": "100%",
         "highlightBackground": false,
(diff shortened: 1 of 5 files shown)


zoltanbedi (commented):
I understand this adds a new option that makes it possible to show the data source name explicitly. Do you know what was the reason that it is not showed in the linked issue?

ovdgInitworks:
I think because the "targets" attribute in the "panelOptions" object was removed/is undefined, so it was never shown.

zoltanbedi:
@ovdgInitworks wouldn't it make sense to try to fix the old behavior?

ovdgInitworks:
That issue was created by a colleague of mine, and both of us just wanted an option and control over the viewing of the data source, instead of requiring multiple targets.

zoltanbedi:
@ovdgInitworks does it make sense for you to show the data source name although they are going to be the same in the list?

ovdgInitworks:
yes

jzinkweg:
Newer versions of grafana have moved the panel options to a separate attribute and no longer provide the full set of panel attributes to the component, as a result, `targets` is no longer available to the component.

We use a single panel with dozens of zabbix servers (one per customer) as sources, so the datasource name is valuable data as it shows which environment (customer) the alert is about.

zoltanbedi:
> Newer versions of grafana have moved the panel options to a separate attribute and no longer provide the full set of panel attributes to the component, as a result, `targets` is no longer available to the component.
> 
> We use a single panel with dozens of zabbix servers (one per customer) as sources, so the datasource name is valuable data as it shows which environment (customer) the alert is about.

Thanks for the explanation @jzinkweg that makes sense.

zoltanbedi:
Sorry we have some issues with CI. Opened a new PR with your changes here https://github.com/grafana/grafana-zabbix/pull/2007

PR #2006, opened 2025-04-07
Fix: Variable item not filtering for item tag

### LINKED ISSUE #1329: "Item tag" option in query variables doesn't work
When using Zabbix 5.4 as data source and when defining a variable of the type 'Query', the query option 'Item tag' doesn't work.
You can fill in whatever you want or you can leave it empty, it doesn't make a difference. The field just isn't taken into account.

The way it should work is like 'Application' did before Zabbix 5.4.

**Software versions**

|8.1.5     | 5.4     | 4.2.4                         |

image

### LINKED ISSUE #1543: Variables based on Zabbix data source, Item tag does not filter
**The problem**
In a dashboard, I tried to create a variable based on a Query for a Zabbix data source.  The Query type is Item. 
Filtering based on the Group and Host work fine. But Item tag doesn't do anything. 

**Expected behavior**
I expected it to filter like it does in the query inside panels, so it limit the number of items displayed.

**Screenshots**
First with the right tag
Capture d’écran 2022-12-21 à 18 06 29

Second with non existing tag.
Capture d’écran 2022-12-21 à 18 08 51

It's the same result if I leave it blank.

**Json model**
The query for the variable looks like :
      "query": {
          "application": "",
          "group": "xxxxx",
          "host": "/.*/",
          "item": "/(?=VF)/",
          "itemTag": "Randomqkflnkjqnef",
          "queryType": "item"
        },

The query in a panel look like :
          "group": {
            "filter": "XXXX"
          },
          "hide": false,
          "host": {
            "filter": "/.*/"
          },
          "item": {
            "filter": "(?=VF)"
          },
          "itemTag": {
            "filter": "$Variable"
          },

**Software versions**
| 9.3  | 6.2.3  | 4.2.10               |

Thx for your help & thx for you nice plugin !

### LINKED ISSUE #1930: Unable to filter the items with item tag at variable level
**Describe the bug**
Hi @alexanderzobnin , i have a variable, of which the type is item, and i enter the proper item tag(as shown below screenshot) to filter but i see all the items without that tag also.

Image

**Expected behavior**
It should show only the items where the tag exists

**Software versions**

|   9.3.6 | 6.0  | 4.4.3                 |

zoltanbedi:
Fixes #1930
Fixes #1543 
Fixes #1329 

TODO

- [x] Add test


itsmylife (approved):
LGTM but I think we need a test as you pointed in the todo list.

PR #2021, opened 2025-05-05
Remove id-token permission for compatibility workflows

zoltanbedi:
The read value for id-token is not available. This PR fixes that.


zoltanbedi:
@itsmylife does this make sense?

itsmylife (approved):

PR #2064, opened 2025-08-03
feat: add per-user authentication

### LINKED ISSUE #2016: Enable per-user authentication in Zabbix plugin based on Grafana user identity
I'm often frustrated when all Grafana users are forced to use the same Zabbix API token for data source queries, regardless of their identity. This limits per-user access control and auditability, especially in environments with centralized authentication like LDAP.

I would like the Zabbix plugin for Grafana to support per-user API token authentication, dynamically tied to the logged-in Grafana user. When a user logs into Grafana, the plugin should:

Use the Grafana username (e.g., from LDAP) to look up the corresponding user in Zabbix.

If the user exists in Zabbix:

Check for an existing API token.

If a token exists, use it.

If not, generate a new token using an admin-level authenticator token.

Use the resulting API token to perform data queries in Zabbix as that user.

If the user does not exist in Zabbix, the data source should not work for that user, and a proper error message should be returned.

Sharing a single service account API token for all users, which prevents proper access control.

Manually mapping API tokens to users, which is unmanageable at scale.

Creating a proxy or middleware to handle user mapping, which adds unnecessary complexity and maintenance overhead.

**Additional context**
This feature would allow Grafana to seamlessly integrate with Zabbix in LDAP-authenticated environments, enabling fine-grained, auditable access control. It aligns with the principle of least privilege and enhances security and user accountability.

christos-diamantis:
This PR introduces per-user authentication to the Grafana Zabbix datasource plugin. When enabled, Grafana users are mapped to corresponding Zabbix users, and all API requests are performed using the permissions of the mapped Zabbix user. This allows organizations to fully leverage Zabbix RBAC and audit capabilities directly from Grafana.

### Key Features
- **Per-User Authentication Toggle:**
New option in the datasource config to enable/disable per-user authentication.
- **Identity Field Selection:**
Admins can choose whether to map users by Grafana username or email.
- **Automatic Zabbix Token Management:**
The plugin will generate and use Zabbix API tokens for each user session, ensuring secure and isolated access.
- **Graceful Error Handling:**
If a Grafana user does not exist in Zabbix, a clear error is shown and access is denied.
- **Backward Compatibility:**
When per-user authentication is disabled, the plugin continues to use the global Zabbix credentials as before.

### How It Works
1. When per-user authentication is enabled, each Grafana request extracts the current user’s identity (username or email).
2. The plugin queries Zabbix for a matching user.
3. If found, a Zabbix API token is generated and used for all API calls in that session.
4. If not found, the user is denied access with a clear error message.

### Configuration
- Enable **Per-user authentication** in the datasource settings.
- Select the identity field (**username** or **email**) for user mapping.
- Ensure each Grafana user has a corresponding Zabbix user.

### Testing
- Integration and unit tests have been added to cover the new authentication flow.
- The devenv bootstrap script now creates a test user for integration testing.

### Documentation
The README and in-app tooltips have been updated to explain the new feature and its requirements.

Feel free to test thoroughly and propose or do any changes if needed.

Closes #2016


zoltanbedi:
@christos-diamantis thanks for your PR. I'll take a look at it shortly.

christos-diamantis:
any update?

christos-diamantis:
any update on this?

pkg/datasource/datasource.go
```diff
@@ -3,8 +3,9 @@ package datasource
 import (
 	"context"
 	"errors"
-	"fmt"
```
matwijec: fmt is needed due https://github.com/grafana/grafana-zabbix/blob/9f66144618f344ef5c01c51d70d822f0d5dcb9f4/pkg/datasource/datasource.go#L121

src/datasource/components/ConfigEditor.tsx
```diff
@@ -380,6 +413,101 @@ export const ConfigEditor = (props: Props) => {
...
+              <Field
+                label={
+                  <Label>
+                    <EditorStack gap={0.5}>
+                      <span>User identity field</span>
+                    </EditorStack>
+                  </Label>
+                }
+              >
+                <Select
```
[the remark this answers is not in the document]
christos-diamantis: Changed the Select to Combobox. Thanks for pointing out.

christos-diamantis:
> > @matwijec any update?
> 
> thanks for the fix. I just tried this feature (and encountered pointed errors) but I'm not a maintainer

Sorry, thought you were a maintainer! What errors did you encountered however? Can you please attach errors and screenshots?

christos-diamantis:
any update?

skaratsolis:
+1

itsmylife:
@christos-diamantis Thanks for this PR. I'll review it today. As it's a big PR it'll take some time but I'm planning to share a feedback today. 
in the meantime could you please merge latest `main` in your branch? That should resolve the conflicts. Also we need all commits to be signed. Otherwise CI won't be happy and I won't be able to merge your PR. 

Please see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

You also have to upload your signing key to GitHub account (Settings → SSH and GPG keys → New GPG key). Without this step, the commits will be locally signed but GitHub will still not show the "Verified" badge which is what we need before merging.

Exact steps:

```bash
# Verify GPG is set up

# Check if a signing key is already configured
git config --global user.signingkey

# If empty, find the key ID from the keyring
gpg --list-secret-keys --keyid-format LONG

# Copy the key ID (the long hex after "sec rsa4096/")
# Set it
git config --global user.signingkey <YOUR_KEY_ID>
git config --global commit.gpgsign true

# Checkout the branch 
git checkout fix/prometheus-series-limit-blur

# Re-sign all commits on your branch since it diverged from main
git rebase -S origin/main

# verify the signature
git log --show-signature -n 2

# force push to update the PR 
git push --force-with-lease origin add-authentication-passthrough
```

pkg/cache/token_cache.go
```diff
@@ -0,0 +1,57 @@
...
+	tc.tokens.Range(func(key, value interface{}) bool {
+		tokenInfo := value.(*TokenInfo)
+		if time.Now().After(tokenInfo.ExpiresAt) {
+			tc.tokens.Delete(key)
+			count++
+		}
+		return true
+	})
+	return count
+}
```
itsmylife, line 74:
```suggestion

}
```

pkg/datasource/auth.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	ds.logger.Info("Per-user authentication successful", "user", identity, "zabbixUser", userName, "tokenCached", true, "ttl", TokenTTL)
+
+	// Cache the token
+	ds.tokenCache.Set(datasourceUID, identity, identity, token, TokenTTL)
+
+	// Scope the user's token to this request only (do not mutate shared instance auth)
+	return zabbixapi.WithPerUserToken(ctx, token), nil
+}
```
itsmylife, line 198:
```suggestion

}
```

pkg/datasource/auth_test.go
```diff
@@ -0,0 +1,200 @@
...
+	responses["user.get"] = `{"result":[]}`
+	s := &settings.ZabbixDatasourceSettings{PerUserAuth: true, PerUserAuthField: "username"}
+	ds, inst := buildTestInstance(t, s, responses, nil)
+
+	ctx := backend.WithUser(context.Background(), &backend.User{Login: "ghost"})
+	_, err := ds.applyPerUserAuth(ctx, inst, "ds-uid")
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "not found in Zabbix")
+}
```
itsmylife, line 229:
```suggestion

}
```

devenv/zabbix60/bootstrap/bootstrap_config.py
```diff
@@ -76,5 +78,27 @@
   except ZabbixAPIException as e:
     print e
 
+print("Creating a user for testing per-user auth")
+groups = zapi.usergoup.get(output="extend", filter={"name": "Zabbix administrators"})
```
itsmylife: there is a typo. It has to be `usergroup`. And this typo exists in other files too.

devenv/zabbix60/bootstrap/bootstrap_config.py
```diff
@@ -76,5 +78,27 @@
...
+if not groups:
+  groups = zapi.usergroup.get(output="extend")
+groupid = groups[0]['usrgrpid']
+
+user_create_params = {
+  "alias": zabbix_test_user,
```
itsmylife: This is deprecated https://www.zabbix.com/documentation/6.0/en/manual/api/reference/user/object

Image

devenv/zabbix60/bootstrap/bootstrap_config.py
```diff
@@ -76,5 +78,27 @@
...
+  "passwd": zabbix_test_pass,
+  "name": "Grafana",
+  "surname": "Test",
+  "type": 1,  # Zabbix user type
```
itsmylife: type is not supported anymore https://blog.zabbix.com/a-guide-to-migrating-to-zabbix-6-0/18569/

`roleid` is required

Image

pkg/datasource/auth.go
```diff
@@ -0,0 +1,142 @@
...
+	token, err := zabbixDS.zabbix.GetAPI().GenerateUserAPIToken(ctx, userId, userName, zabbixVersion)
+	if err != nil {
+		ds.logger.Error("Failed to generate token", "userId", userId, "error", err)
+		return ctx, errors.New("failed to generate Zabbix API token for user: " + err.Error())
+	}
+
+	ds.logger.Info("Per-user authentication successful", "user", identity, "zabbixUser", userName, "tokenCached", true, "ttl", TokenTTL)
+
+	// Cache the token
+	ds.tokenCache.Set(datasourceUID, identity, identity, token, TokenTTL)
```
itsmylife: `identity` used twice. I think it's intentional but don't we need to use userId?

pkg/datasource/auth.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	for _, excludedUser := range exclusionList {
+		if strings.EqualFold(identity, excludedUser) {
+			excluded = true
+			break
+		}
+	}
+
+	if excluded {
+		ds.logger.Info("User is excluded from per-user authentication, using stored credentials", "user", identity)
```
itsmylife: we shouldn't log/leak anything related to user.

pkg/datasource/auth.go
```diff
@@ -0,0 +1,142 @@
...
+		return ctx, nil
+	}
+
+	// Check token cache first
+	if tokenInfo, ok := ds.tokenCache.Get(datasourceUID, identity, identity); ok {
+		ds.logger.Debug("Using cached token", "user", identity, "expiresIn", time.Until(tokenInfo.ExpiresAt).Round(time.Minute))
```
itsmylife: same applies here. No user info in the logs please.

src/datasource/components/ConfigEditor.tsx
```diff
@@ -437,6 +470,104 @@ export const ConfigEditor = (props: Props) => {
...
+                    value: options.jsonData.perUserAuthField || 'username',
+                  }}
+                  onChange={jsonDataSelectHandler('perUserAuthField', options, onOptionsChange)}
+                />
+              </Field>
+
+              {userFetchWarning && (
+                <div style={{ color: 'orange', marginBottom: '10px' }}>
+                  <Icon name="exclamation-triangle" /> {userFetchWarning}
+                </div>
```
itsmylife: we can use InfoBox component instead https://developers.grafana.com/ui/latest/index.html?path=/story/information-deprecated-infobox--feature-info-box&args=severity:warning

.changeset/curvy-lizards-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'grafana-zabbix': major
```
itsmylife: We can mark this as `minor`

pkg/datasource/datasource.go
```diff
@@ -39,10 +41,26 @@ type ZabbixDatasourceInstance struct {
...
+	go ds.startTokenCleanup()
+
+	return ds
+}
+
+func (ds *ZabbixDatasource) startTokenCleanup() {
+	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
+
+	for range ticker.C {
```
itsmylife: `startTokenCleanup` starts a ticker goroutine that we can't stop. For production it is ok since `NewZabbixDatasource` is a singleton, but it leaks one goroutine per call in tests (in datasource_test.go we create it inside a loop). It's also not a big issue but still want a better thing. 

Maybe we can pass a context + interval and select on ctx.Done(), then expose a small Close() so tests can stop it? Something like:

```go
func (ds *ZabbixDatasource) startTokenCleanup(ctx context.Context, interval time.Duration) {
	ticker := time.NewTicker(interval)
	defer ticker.Stop()
	for {
		select {
		case <-ctx.Done():
			return
		case <-ticker.C:
			// cleanup
		}
	}
}
```

itsmylife (changes requested):
Overall it's a good implementation thank you for sparing time and creating this PR and your patience. 

There are a few linting issues which you can catch by running `yarn lint` for frontend. I've pointed the backend problems too.

devenv/zabbix60/bootstrap/bootstrap_config.py
christos-diamantis: fixed!

devenv/zabbix60/bootstrap/bootstrap_config.py
christos-diamantis: resolved

devenv/zabbix60/bootstrap/bootstrap_config.py
christos-diamantis: indeed. good catch! also "user_groups" is not valid anymore for `user.create` method. it must be `usrgrps`:
https://www.zabbix.com/documentation/6.0/en/manual/api/reference/user/create
fixed on latest push

pkg/datasource/auth.go
christos-diamantis: Good catch! The cache key has to be `datasourceUID` and `identity` because the lookup runs before we query Zabbix, so `userId` isn't known yet. 
I've cleaned it up though: dropped the redundant arg, and `Set` now stores the real userId in `TokenInfo.UserID` instead of duplicating identity.

pkg/datasource/auth.go
christos-diamantis: fixed on latest commit

pkg/datasource/auth.go
christos-diamantis: fixed on latest commit

src/datasource/components/ConfigEditor.tsx
christos-diamantis: actually InfoBox is showing deprecated on the link you referenced:

`Deprecrated component, use Alert with info severity`

changed it to Alert with severity "warning"

.changeset/curvy-lizards-report.md
christos-diamantis: done

pkg/datasource/datasource.go
christos-diamantis: Nice catch. Wrapped the cleanup ticker in a `select` on `ctx.Done()` and added a `Close()` that cancels it. `NewZabbixDatasource` now passes a cancellable context + interval into `startTokenCleanup`, and the tests defer `ds.Close()` so the goroutine no longer leaks per iteration. Production is unchanged (the singleton just never calls `Close()`).

christos-diamantis:
> Overall it's a good implementation thank you for sparing time and creating this PR and your patience.
> 
> There are a few linting issues which you can catch by running `yarn lint` for frontend. I've pointed the backend problems too.

Thank you for the review! I have tackled all the issues and pushed. Awaiting for your feedback.

pkg/datasource/auth.go
```diff
@@ -0,0 +1,143 @@
...
+		identity = user.Email
+	default:
+		identity = user.Login
+	}
+
+	// If identity is empty, skip per-user auth
+	if identity == "" {
+		ds.logger.Debug("User identity is empty, skipping per-user auth")
+		return ctx, nil
+	}
```
itsmylife, line 60: Just a question: Should this error out?

devenv/zabbix70/docker-compose.yml
```diff
@@ -91,6 +91,8 @@ services:
       ZBX_API_URL: http://zabbix-web:8080
       ZBX_API_USER: Admin
       ZBX_API_PASSWORD: zabbix
+      ZBX_TEST_USER: grafana_test
+      ZBX_TEST_PASS: grafana_test_pass
```
itsmylife: When I run docker I'm getting this error 
```
Creating a user for testing per-user auth
Failed to create user: Invalid params. Incorrect value for field "/1/passwd": must not contain user's name, surname or username.
```
password has to be a different value than username.

src/datasource/types/config.ts
```diff
@@ -33,6 +33,9 @@ export type ZabbixDSOptions = {
    * */
   dbConnectionDatasourceId?: number;
   schema?: number;
+  perUserAuth?: boolean;
+  perUserAuthField?: 'username' | 'email';
```
itsmylife, line 37: username is working but not email. Zabbix stores email under `medias[].sendto`, not a direct email property. Maybe that's the reason?

itsmylife (changes requested):
Thanks a lot for taking quick actions. 

I run some additional manual tests and some tests with an agent. I'm testing this PR throughly because of it's size and importance. Thank you for your patience.

Here is the result of agent assisted test, I use zabbix 7.0 devenv:

Testing environment details:
In Zabbix
Create host groups named Only-A and Only-B.
Put at least one host in each group.
Create these user groups:
Grafana-A, with read access only to Only-A.
Grafana-B, with read access only to Only-B.
Create these Zabbix users:
user-a, assigned to Grafana-A.
user-b, assigned to Grafana-B.
Ensure their roles permit API access.
In Grafana, create matching Viewer users named user-a and user-b.

**Concurrent token generation**
I restarted Grafana to clear the in-memory token cache, then sent 20 authenticated requests concurrently through the zabbix-testing datasource.

Results:
All 20 requests logged Authenticating user with Zabbix.
All 20 independently generated a token for the same Zabbix user.
6 requests succeeded.
14 requests failed with HTTP 500: Invalid params. Not authorized.
This happens because the cache check and token generation are not synchronized. Each token.generate call invalidates the token generated by another concurrent request.

Suggested action:
Deduplicate token generation per datasource/user, for example with singleflight or a keyed lock.
Recheck the cache after acquiring the lock.
Only one request should call token.generate; waiting requests should use the resulting cached token.

**Invalid cached token recovery**
I first made a successful request as user-a, which generated and cached a valid token. I then regenerated Zabbix-Grafana-Token_user-a directly in Zabbix, invalidating Grafana’s cached value.

I sent five cache-busted requests afterward.

Results:
All five returned HTTP 500: Invalid params. Not authorized.
Every request logged Using cached token.
Retrying did not generate a replacement token.
After restarting Grafana, the next request generated a new token and succeeded with HTTP 200.
This confirms that rejected tokens remain cached for the full 24-hour TTL.

Suggested action:
When Zabbix rejects a per-user token as unauthorized, remove that user’s token from tokenCache.
Retry token generation once.
Do not fall back to the shared Admin credentials.
Guard the regeneration with the same per-user synchronization used to fix test 5.
Limit it to one retry to avoid loops.
One additional observation: repeating the exact same query initially returned previously cached successful query data without contacting Zabbix. I varied the query parameters to bypass the response cache and expose the invalid token. Per-user response caching should therefore also be partitioned by identity/token or disabled for per-user requests.

pkg/datasource/auth.go
christos-diamantis: Ideally we should make it error out. The whole purpose for this is to respect the security model of Zabbix, so when per-user auth is enabled we should strictly follow that. Gonna tackle it

devenv/zabbix70/docker-compose.yml
christos-diamantis: You are right. Completely forgot this caveat. Will change to something else.

src/datasource/types/config.ts
christos-diamantis: actually, the email selection is only for when the users in zabbix are using SAML authentication with any known provider (e.g. Azure). On most cases, the username of the user in Zabbix will be his email, not the UPN and grafana actually stores that to email.

christos-diamantis:
> Thanks a lot for taking quick actions.
> 
> I run some additional manual tests and some tests with an agent. I'm testing this PR throughly because of it's size and importance. Thank you for your patience.
> 
> Here is the result of agent assisted test, I use zabbix 7.0 devenv:
> 
> Testing environment details: In Zabbix Create host groups named Only-A and Only-B. Put at least one host in each group. Create these user groups: Grafana-A, with read access only to Only-A. Grafana-B, with read access only to Only-B. Create these Zabbix users: user-a, assigned to Grafana-A. user-b, assigned to Grafana-B. Ensure their roles permit API access. In Grafana, create matching Viewer users named user-a and user-b.
> 
> **Concurrent token generation** I restarted Grafana to clear the in-memory token cache, then sent 20 authenticated requests concurrently through the zabbix-testing datasource.
> 
> Results: All 20 requests logged Authenticating user with Zabbix. All 20 independently generated a token for the same Zabbix user. 6 requests succeeded. 14 requests failed with HTTP 500: Invalid params. Not authorized. This happens because the cache check and token generation are not synchronized. Each token.generate call invalidates the token generated by another concurrent request.
> 
> Suggested action: Deduplicate token generation per datasource/user, for example with singleflight or a keyed lock. Recheck the cache after acquiring the lock. Only one request should call token.generate; waiting requests should use the resulting cached token.
> 
> **Invalid cached token recovery** I first made a successful request as user-a, which generated and cached a valid token. I then regenerated Zabbix-Grafana-Token_user-a directly in Zabbix, invalidating Grafana’s cached value.
> 
> I sent five cache-busted requests afterward.
> 
> Results: All five returned HTTP 500: Invalid params. Not authorized. Every request logged Using cached token. Retrying did not generate a replacement token. After restarting Grafana, the next request generated a new token and succeeded with HTTP 200. This confirms that rejected tokens remain cached for the full 24-hour TTL.
> 
> Suggested action: When Zabbix rejects a per-user token as unauthorized, remove that user’s token from tokenCache. Retry token generation once. Do not fall back to the shared Admin credentials. Guard the regeneration with the same per-user synchronization used to fix test 5. Limit it to one retry to avoid loops. One additional observation: repeating the exact same query initially returned previously cached successful query data without contacting Zabbix. I varied the query parameters to bypass the response cache and expose the invalid token. Per-user response caching should therefore also be partitioned by identity/token or disabled for per-user requests.

Thank you very much for the detailed test! I will take a look later on today and try to tackle these issues.

devenv/zabbix70/docker-compose.yml
christos-diamantis: Changed ZBX_TEST_PASS to a policy-compliant value in all four devenvs (compose files + bootstrap Dockerfile defaults).

pkg/datasource/auth.go
christos-diamantis: an empty identity now fails the request with `per-user authentication is enabled but the Grafana user's <field> is empty` instead of silently proceeding with the stored credentials. Anonymous access and empty identity are now handled the same strict way, so enabling per-user auth always enforces the Zabbix permission model. Added tests for both the username and email mapping cases.

christos-diamantis:
@itsmylife fixed the issues

**Concurrent token generation**: token resolution now goes through a singleflight keyed by datasource:identity, with a cache re-check inside the flight. Only one request generates; the rest wait and reuse the result. A test with 20 concurrent cold-cache requests asserts exactly one token.generate call and that all requests end up with the same token.

**Invalid cached token recovery**: when Zabbix rejects a per-user token, the request layer now evicts it from the cache (compare-and-delete, so a concurrently regenerated fresh token is never evicted by a straggler), regenerates once through the same singleflight, and retries the request with the new token. The refresher is cleared on the retry, so a second rejection surfaces the error instead of looping — and there is still no fallback to the stored credentials. Covered by a refresh-and-retry test and a no-loop test.

**Response cache partitioning**: the API response cache key is now scoped by the per-user token, so responses fetched under one user's permissions are never served to another user (or to the shared-credential path, which keeps its own partition and existing behavior). Covered by a test asserting user-A's cached response is not reused for user-B or for shared-credential requests.

itsmylife:
@christos-diamantis Can you handle the conflicts? And then I think we can merge this PR.

christos-diamantis:
@itsmylife Done. Conflicts resolved.

christos-diamantis:
@itsmylife  I just checked and realized the integration tests i had written were done in the beggining of the PR and are silently skipping testing the feature. writing a commit right now to fix this. :)

christos-diamantis:
@itsmylife Done.

- fixed integration tests which were silently failing instead of raising error
- Bootstrap: construct the API client inside the retry loop - the
  constructor performs a version check that fails while the Zabbix
  database is still initializing
- Add handler-level smoke tests: full ZabbixAPIHandler flow with a mock
  Zabbix server, asserting the data query carries the per-user token
  (guards against the handler dropping the auth context) and that a
  missing Zabbix user yields 403 with no stored-credential fallback
- Assert bootstrap calls (user lookup, token management) always use the
  stored session in the recovery tests
- Cover TokenCache.CompareAndDelete's conditional-eviction semantics
- Add ConfigEditor tests for the per-user auth section (visibility and
  the permissions warning)

itsmylife:
@christos-diamantis some checks are failing. 
```
 Container zabbix60-bootstrap-1  Started
--- bootstrap logs ---
  File "/bootstrap_config.py", line 30
    except:
         ^
SyntaxError: invalid syntax
bootstrap exited with code 1
```

christos-diamantis:
> @christos-diamantis some checks are failing.
> 
> ```
>  Container zabbix60-bootstrap-1  Started
> --- bootstrap logs ---
>   File "/bootstrap_config.py", line 30
>     except:
>          ^
> SyntaxError: invalid syntax
> bootstrap exited with code 1
> ```

fixed and validated locally

christos-diamantis:
failed again. integration test for 6.0 is the only one running on python 2.7. there was a fix i did on the bootstrap of 70/72/74 and assumed was not needed for 60. but turns out it was needed.

itsmylife:
@christos-diamantis Looks like CI is happy so far. 🤞 for the e2e tests

itsmylife (approved):
Thank you for this huge work and your time. It's amazing. Thank you for your swift replies. 🚀

PR #2065, opened 2025-08-06
Forward refID to resultant dataframe

### LINKED ISSUE #1982: query return data in "wide" format without refid
When i'll trying to select only 1 item for N hosts in query , zabbix was return data in "wide" format without refid. 
In this case, i can't use filter to transform data steps only for "wide" format query (it was applying for all Queries forcedly e.g A,B,C).
Data Alignment disabled in plugin settings (alexanderzobnin-zabbix-app plugin version 5.0.2, Zabbix API version: 6.0.10, grafana version 11.1.3).

How can i fix this behavior? Need to take data from zabbix in format "1 host = 1 table" AND\OR with refid (without using regexp for taking few various metrics per time)

Image

Image

Image

gelicia:
Fixes #1982

With a refID of `test test`
Screenshot 2025-08-06 at 4 11 41 PM

Before
Screenshot 2025-08-06 at 4 18 40 PM

After
Screenshot 2025-08-06 at 4 11 33 PM

### DIFF
--- a/src/datasource/responseHandler.spec.ts
+++ b/src/datasource/responseHandler.spec.ts
@@ -6,6 +6,7 @@ describe('convertToWide', () => {
     let frames = convertToWide([
       new MutableDataFrame({
         name: 'SLI',
+        refId: 'A',
         fields: [
           { name: TIME_SERIES_TIME_FIELD_NAME, values: [1], type: FieldType.time },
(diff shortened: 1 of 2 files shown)


zoltanbedi (approved):
Thanks @gelicia!

Issue #2061, opened 2025-07-28
More info about acknowledges from zabbix


HH-Harry:
Hello, we are looking for more information from Zabbix about acknowledge actions. Currenty, only comments are imported, so if I do any "update" action in zabbix without comment (acknowledge, unacknowledge, change severity, etc.), no information about "what was done" is shown in Grafana, only user name.

Example:
What I see in zabbix:
<img width="512" height="156" alt="Image" src="https://github.com/user-attachments/assets/80f43a57-16a7-48e5-b389-6af830fd4540" />
What I see in Grafana (I also notice there is a problem in formatting/line ordering, when there is no comment, but this should be solved by this request):
<img width="397" height="107" alt="Image" src="https://github.com/user-attachments/assets/a0d96736-4a53-4423-b11a-b4cda3e586d9" />

My idea is to add description of action at the beginning of imported comment (inside brackets). Most important for me is info about acknowledge / unacknowlegde, but I expect other action should be done the same way.

Example (based on data from previous) of what I want to see:
<img width="450" height="111" alt="Image" src="https://github.com/user-attachments/assets/132a2c97-e2a7-4b4e-83ba-d719807ded80" />

zoltanbedi:
Hey @HH-Harry, thanks for raising this issue for us. We are accepting PRs, feel free to open one with your proposed changes. I think those make sense.

HH-Harry:
In the end, I tried :) PR created, I am just not sure if it is correct, so rest on you.

zoltanbedi:
Fixed in 2071

PR #2071, opened 2025-08-24
More info about acknowledges from zabbix

HH-Harry:
This PR is trying to add functionality requested in [#2061 More info about acknowledges from zabbix](https://github.com/grafana/grafana-zabbix/issues/2061)

### Key features

- already described in [Enhancement request](https://github.com/grafana/grafana-zabbix/issues/2061)

### How It Works

- using bitwise AND checks of [**action** field in zabbix event.acknowledges](https://www.zabbix.com/documentation/current/en/manual/api/reference/event/acknowledge) keywords are added at beginning of ack.message field on problem panel in grafana in fllowing order:
  - (un)acknowledged
  - (un)supressed
  - changed severity

### Testing
- No testing was done, sorry


zoltanbedi:
I took a look and I think what would make sense is to have those icons or something like that as a new column just like in zabbix.

HH-Harry:
That would be great, but I am unable to do something like that :(
So as a first step, I look at least to have this text info

zoltanbedi:
@HH-Harry can you please accept the cla again?

zoltanbedi (approved):
Awesome job! thanks for this PR.

PR #2090, opened 2025-09-24
Refactor regex pattern validation to use timeout-based approach

KristianGrafana:
- Remove isPathologicalRegex function and replace with MatchTimeout
- Simplify parseFilter by relying on runtime timeout protection
- Add comprehensive timeout test for pathological regex patterns
- Set 5-second timeout for all compiled regex operations


zoltanbedi (approved):
LGTM

PR #2110, opened 2025-10-21
Fix: ensure that applicationids parameter only gets passed when the datasource supports it.

jcolladokuri:
Zabbix 5.0.x supported filtering `Problems` feature with `applications`. When this got removed, we removed the filter dropdown from the UI, but failed to check whether applications were supported before sending out the request with the parameters.

This was causing dashboards that had been created with zabbix version `5.0.x` to fail when querying with newer versions of our plugin with error: `Invalid params. Invalid parameter "/": unexpected parameter "applicationids".`

These changes now ensure that we also check whether applications filter should be supported before sending the backend request to fetch problems.

How to test:
- use the attached JSON file. This was  created using zabbix50 and applying an `applicationids` filter for `Problems` query type OR 

- run the `zabbix50` test environment:
    ```
       cd devenv/zabbix50
       docker-compose up -d
    ```
- create a dashboard that queries for `Problems` and filters with applications then export the dashboard JSON
- stop the `zabbix50` test environment and start the `zabbix74` test environment
    ```
          docker-compose stop
          cd ../zabbix74
          docker-compose up -d
     ```
- import the dashboard you created above, it should load and work as expected.

Bottom panel was created using zabbix50 and it used the application filter. Both panels now load as expected:
Screenshot 2025-10-21 at 2 28 25 PM

Fixes https://github.com/grafana/grafana-zabbix/issues/1852

### DIFF
--- a/.changeset/cuddly-cloths-join.md
+++ b/.changeset/cuddly-cloths-join.md
@@ -1,0 +1,5 @@
+---
+'grafana-zabbix': minor
+---
+
+Fix support of applicationids filters with Zabbix problems for versions older than 5.0.x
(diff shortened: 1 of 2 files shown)


jcolladokuri:
Added the change set :)! Thanks for reviewing

zoltanbedi (approved):

PR #2113, opened 2025-10-24
Add the ability to convert specific tags to columns on problems panel

### LINKED ISSUE #2112: Add the ability to convert specific tags to columns on problems panel
A lot of times, we need to be able to see specific tags really quickly on the problems panel and its impossible, since we have to open the problem details to see the exact tag value.

We need to be able to specify one or more tags to be converted to columns.
Also if a tag is present multiple times, it should show all the values seperated

christos-diamantis:
This closes #2112

Added the ability on problems panel to specify a comma separated list of tag names to be converted to columns.
If a tag name is present multiple times, it will return the value of all tags separated with comma.
For optimum readability, the tag names are Capitalized for the visible column name. 
Also, for optimum readability, the custom tags are always placed before the "Tags" column.

In case a tag is not there for a problem, an empty string is returned.

### DIFF
--- a/devenv/zabbix74/bootstrap/Dockerfile
+++ b/devenv/zabbix74/bootstrap/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13-slim-bookworm@sha256:0c5171fd1e80d3133604e1006aa5f788c5f020631537dd1e09edcbe874bb8192
+FROM python:3.13-slim-bookworm@sha256:97e9392d12279f8c180eb80f0c7c0f3dfe5650f0f2573f7ad770aea58f75ed12
 
 ENV ZBX_API_URL=http://zabbix-web:8080
 ENV ZBX_API_USER="Admin"


src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -72,6 +72,37 @@ export const ProblemList = (props: ProblemListProps) => {
...
 
+    const capitalizeFirstLetter = (str: string) => str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
+
+    const customTagColumns =
+      panelOptions.customTagColumns && panelOptions.customTagColumns.trim().length > 0
+        ? panelOptions.customTagColumns
+            .split(',')
+            .map((t) => t.trim())
+            .filter(Boolean)
+            .map((tagName) =>
```
jcolladokuri: this can be simplified to a single `.map` function call and inside that call we can trim the tag name and check if it is empty, so something like:

```
          panelOptions.customTagColumns
            .split(',')
            .map((tagName) => {
              const trimmed = tagName.trim();
              if (trimmed.length > 0) {
                columnHelper.accessor(...)
              }
            })
```

src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -72,6 +72,37 @@ export const ProblemList = (props: ProblemListProps) => {
   const columns = useMemo(() => {
     const highlightNewerThan = panelOptions.highlightNewEvents && panelOptions.highlightNewerThan;
 
+    const capitalizeFirstLetter = (str: string) => str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
+
+    const customTagColumns =
```
jcolladokuri: Given that `customTagColumns` has a bit more complex logic, can we please move it to its own helper function?

Additionally, it would be great if we could create a `utils.ts` file and add that helper function there along with `capitalizeFirstLetter` and add some unit tests for both :)

jcolladokuri (changes requested):
Thanks so much for your contribution! Everything looks and works great, just a couple minor things to address :)

Additionally, there has been a few changes on `main` so please be sure to sync your branch to latest main, and address the lint errors that were called out by the Github action.

christos-diamantis:
Hello @jcolladokuri! Thank you very much for your review and input.
I transfered the `capitalizeFirstLetter` function to utils.ts and also created a `parseCustomTagColumns` function.

on `Problems.tsx` I created a `buildCustomTagColumns` function that needs React + JSX and could not offload to `utils.ts`.

So the final logic is the following:

- parsing, sanitization, data formatting on utils.ts
- react, tsx, tanstack on Problems.tsx

src/panel-triggers/components/Problems/Problems.tsx
christos-diamantis: I offloaded the parsing, sanitization and data formatting on utils.js and kept the tsx and columns building on the Problems.tsx

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grafana-zabbix",
-  "version": "6.0.3",
+  "version": "6.1.0",
```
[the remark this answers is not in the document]
christos-diamantis: I confused the `yarn changeset` with `yarn changeset version`. I have ran it

PR #2120, opened 2025-11-14
Move health check to the backend

jcolladokuri:
This PR moves the health check to backend only leaving in the frontend the functionality to test the dbconnector datasource. 

Leaving the `dbconnector.testDataSource` should be fine since the datasource types we allow for db connection with Zabbix already are backend datasources, and so their health requests would go through the backend.

Verified:
Clicking test and seeing a `health` request go out.

IMPORTANT: While testing this in the UI, I found a bug with the config editor - whenever a change is made in the UI and tested, the changes don't take effect (i.e. disabling trends, keeps `trends` set to `true`, enabling db connection keep `dbConnectionEnabled` set to `false` and so on.).  Created a separate [issue](https://github.com/orgs/grafana/projects/457/views/40?pane=issue&itemId=3627315751&issue=grafana%7Coss-big-tent-squad%7C132) to fix this

Fixes https://github.com/grafana/oss-big-tent-squad/issues/124
Fixes https://github.com/grafana/grafana-zabbix/issues/2004


src/datasource/zabbix/zabbix.ts
```diff
@@ -208,32 +208,21 @@ export class Zabbix implements ZabbixConnector {
    * ```
    */
   testDataSource() {
```
itsmylife: I don't think we need this anymore. I think when we hit `save&test` button in the config editor it just calls `CheckHealth` endpoint.

itsmylife (commented):
It looks good but we need to double check if we still need `testDatasource()`

src/datasource/zabbix/zabbix.ts
jcolladokuri: This one we need to be able to fetch the dbConnection status when a datasource is selected for `dbConnectionEnabled`

PR #2133, opened 2025-12-03
Fix always fetch Zabbix version before issuing new requests

jcolladokuri:
Previously we were only fetching the version when the version was `0`. This generally worked, but posed some problems when customers were updating their Zabbix version, specifically when upgrading from a version < `7.2.x` to `7.2.x` or above.

Before `7.2.x`, an `auth` parameter was still supported when issuing a zabbix request, this was deprecated in `6.4.x` and later removed in `7.2.x`. When a user was on a version < `7.2.x` all the outgoing requests would add this `auth` parameter. When upgrading to `7.2.x` this was  a problem, because the version was not `0`, hence, not requiring getting the version again, but also because we were still building the request considering an older version and adding the `auth` parameter, when this was no longer supported.

This PR removes the check for `version == 0`, though this now means that every request that goes out will check the version before hand, I think this will give us a more accurate representation of the version that needs to be used.

fixes https://github.com/orgs/grafana/projects/457/views/40?pane=issue&itemId=3683181283&issue=grafana%7Coss-big-tent-squad%7C135

### DIFF
--- a/pkg/zabbix/zabbix.go
+++ b/pkg/zabbix/zabbix.go
@@ -54,7 +54,6 @@ func (ds *Zabbix) Request(ctx context.Context, apiReq *ZabbixAPIRequest) (*simpl
 		ds.logger.Error("Error querying Zabbix version", "error", err)
 		ds.version = -1
 	} else {
-		ds.logger.Debug("Got Zabbix version", "version", version)
 		ds.version = version
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/zabbix/zabbix.go
```diff
@@ -48,16 +48,14 @@ func (zabbix *Zabbix) GetAPI() *zabbixapi.ZabbixAPI {
...
-			ds.logger.Debug("Got Zabbix version", "version", version)
-			ds.version = version
-		}
+	
+	version, err := ds.GetVersion(ctx)
+	if err != nil {
+		ds.logger.Error("Error querying Zabbix version", "error", err)
+		ds.version = -1
+	} else {
+		ds.logger.Debug("Got Zabbix version", "version", version)
```
zoltanbedi: This debug log is going to flood us with this version logs. I think we can remove this log and add the version to other logs as an attribute.

PR #2140, opened 2025-12-17
Adds support for host tags

jcolladokuri:
## Sumary
When dealing with multiple hosts, it can be hard for customers filter through and figure out which host to query metric data from. This PR aims to make this easier by adding support for host tags so that there is another layer of filtering / grouping applied for hosts. 

## Detailed explanation
- Adds new UI components to allow adding one or more host tag filter, and a switch to choose between `AND/OR` and `OR` operators when using more than one filter following Zabbix's UI:
  
    https://github.com/user-attachments/assets/c971f5eb-7e93-4238-bd6b-902cc657c014

    https://github.com/user-attachments/assets/5f8996de-684e-4ffa-b98e-8e205c4fc1df

- Modifies the existing `getHosts` function to make a call to the backend with a few additional parameters to `extend` (essentially extract) the host tags for a given selected group. No backend changes were required for this.

## Why
To make it easier for customers to query metric data when dealing with multiple hosts.

## How to test
- Go to explore or a dashboard and create a Zabbix query where the query type is `Metrics`
- The easiest way to test is by selecting `/.*/` for Groups, checking the returned `Hosts` they should all be there
- Add a host tag filter and change the keys and operators as well as switching from `AND/OR` to `OR` you should see how the values returned for `Host` changes

## Future work
Adding variable support for host tags once this is completed.

Fixes: https://github.com/orgs/grafana/projects/457/views/40?pane=issue&itemId=3609900134&issue=grafana%7Coss-big-tent-squad%7C126 and https://github.com/grafana/grafana-zabbix/issues/927


src/datasource/components/QueryEditor/utils.ts
```diff
@@ -11,3 +14,27 @@ export const getVariableOptions = () => {
...
 };
+
+export function processHostTags(hosts: Host[]): Tag[] {
+  const hostTags = hosts.map((host) => host.tags || []).flat();
+  // deduplicate tags
+  const uniqueHostTags = _.uniqBy(hostTags, (tag) => `${tag.tag}`);
+  return uniqueHostTags;
+}
+
+/** Get the label for a host tag option
```
itsmylife:
```suggestion
/**
 * Get the label for a host tag option
```

src/datasource/components/QueryEditor/utils.ts
```diff
@@ -11,3 +14,27 @@ export const getVariableOptions = () => {
     label: `$${v.name}`,
   }));
 };
+
+export function processHostTags(hosts: Host[]): Tag[] {
+  const hostTags = hosts.map((host) => host.tags || []).flat();
+  // deduplicate tags
+  const uniqueHostTags = _.uniqBy(hostTags, (tag) => `${tag.tag}`);
```
itsmylife: isn't `tag.tag` is already string? Why are we using backticks?

src/datasource/components/QueryEditor/HostTagQueryEditor.tsx
```diff
@@ -0,0 +1,137 @@
...
+              />
+              {filter.operator !== HostTagOperatorValue.Exists &&
+                filter.operator !== HostTagOperatorValue.DoesNotExist && (
+                  <Input
+                    value={filter.value}
+                    onChange={(evt: FormEvent<HTMLInputElement>) =>
+                      setHostTagFilterValue(index, evt?.currentTarget?.value)
+                    }
+                    width={19}
+                  />
```
itsmylife: A short placeholder would be useful. At first glance I've missed what I need to do with the input field.

itsmylife (commented):
I've pointed mostly the nitpicks. 
But when I apply a weird host tag it still returns data. I'm not sure if that how it should be. 

Image


When I try to do the same in zabbix it doesn't show anything. I think hostTags are not really affecting the query that is sent to the Zabbix. 

Image

jcolladokuri:
@itsmylife thanks for reviewing and for all your feedback! I ended up changing the logic a bit for when we append `/.*/` the problem was that we still had that even when no hosts were returned and that is why the query was returning results.

Now, we only add that if there are hosts results

PR #2152, opened 2025-12-19
Add item count warning in query editor for large result sets

itsmylife:
## Summary

Adds a non-intrusive warning banner in the query editor that alerts users when their query matches a large number of items (>= 500). This helps users understand that their query may return a large amount of data and suggests using more specific filters.

Part of https://github.com/grafana/oss-big-tent-squad/issues/127

## Changes

- Added `ITEM_COUNT_WARNING_THRESHOLD` constant (500 items) in `src/datasource/constants.ts`
- Created new `ItemCountWarning` component in `src/datasource/components/ItemCountWarning.tsx`
- Updated `MetricsQueryEditor` to track and report the count of items matching the current filter
- Integrated the warning component into the main `QueryEditor` component

## How it works

- When items are loaded for the dropdown in the Metrics query editor, the component counts how many items match the current item filter
- If using a regex filter like `/.*/`, it applies the regex to count matching items
- If the count is >= 500, a warning banner appears at the top of the query editor
- The warning is purely informational - queries still execute normally
- The warning only appears for the "Metrics" query type

## Screenshot

The warning appears as a subtle banner with a warning icon:

> I set the limit as 5 just to show the warning

grafik

## Why

Queries that match thousands of items via wildcard filters (e.g., `/.*/`) can return massive amounts of data and potentially overload the Zabbix server. This proactive warning helps users make informed decisions about their query scope without adding friction to the normal query flow.

### DIFF
--- a/src/datasource/components/ItemCountWarning.tsx
+++ b/src/datasource/components/ItemCountWarning.tsx
@@ -1,7 +1,5 @@
 import React from 'react';
-import { css } from '@emotion/css';
-import { GrafanaTheme2 } from '@grafana/data';
-import { Icon, useStyles2 } from '@grafana/ui';
+import { Alert } from '@grafana/ui';
 import { ITEM_COUNT_WARNING_THRESHOLD } from '../constants';
 
 interface ItemCountWarningProps {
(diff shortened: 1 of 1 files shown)


src/datasource/components/ItemCountWarning.tsx
```diff
@@ -0,0 +1,56 @@
...
+  threshold = ITEM_COUNT_WARNING_THRESHOLD,
+}) => {
+  const styles = useStyles2(getStyles);
+
+  if (itemCount < threshold) {
+    return null;
+  }
+
+  return (
+    <div className={styles.warningContainer}>
```
jcolladokuri: Should we use `grafana.ui`'s `Alert` here as well to avoid doing custom styling ourselves?

src/datasource/components/QueryEditor.tsx
```diff
@@ -198,6 +205,7 @@ export const QueryEditor = ({ query, datasource, onChange, onRunQuery }: ZabbixQ
 
   return (
     <Stack direction="column">
+      {queryType === c.MODE_METRICS && <ItemCountWarning itemCount={itemCount} />}
```
jcolladokuri: Similar question here, should this be shown at the bottom of the query editor?

Issue #1331, opened 2021-10-12
failed to receive call resource response: rpc error: code = ResourceExhausted desc = grpc: received message larger than max


cronop:
Intermittent error when running a topN time series query on item utilization across a large number (>1000) of switch interfaces.

Similar to https://github.com/alexanderzobnin/grafana-zabbix/issues/1087

grafana-zabbix 4.2.4 
grafana Version 8.1.6 (commit: 4a4083716c, branch: HEAD)
mysqld  Ver 10.5.12-MariaDB for Linux on x86_64 (MariaDB Server)
zabbix_server (Zabbix) 5.0.15
aws linux  4.14.246-187.474.amzn2.x86_64

grafana.log
lvl=eror msg="Failed to call resource" logger=context userId=1 orgId=1 uname=admin error="failed to receive call resource response: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (17089470 vs. 16777216)"
lvl=eror msg="Failed to call resource" logger=context userId=4 orgId=1 uname=admin error="failed to receive call resource response: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (21246971 vs. 16777216)"

JSON response
Failed to call resource
Object
status:500
statusText:"Internal Server Error"
data:Object
message:"Failed to call resource"
config:Object
url:"api/datasources/2/resources/db-connection-post"
method:"POST"
headers:Object
hideFromInspector:false
data:Object
retry:0
message:"Failed to call resource"

![image](https://user-images.githubusercontent.com/31140497/136971850-555a8e25-bb2c-4744-b9ce-d72952a87318.png)
![Screenshot 2021-10-12 100524](https://user-images.githubusercontent.com/31140497/136971874-767439b8-ebcc-487f-b68a-d94a6422670e.jpg)

iRomanyshyn:
I've got the same error when trying to do a big requests. It looks like some settings can be changed on the remote database side to extend possible size of the message.

cronop:
> 
> 
> I've got the same error when trying to do a big requests. It looks like some settings can be changed on the remote database side to extend possible size of the message.

добрый ранок, any ideas which settings?

iRomanyshyn:
> 
> 
> > I've got the same error when trying to do a big requests. It looks like some settings can be changed on the remote database side to extend possible size of the message.
> 
> добрый ранок, any ideas which settings?
We have no "ы" letter in Ukrainian alphabet, BTW :)

[...]

PR #2210, opened 2026-01-16
Add configurable gRPC message size limits via environment variables

mario-zelger:
This PR adds the ability for users to configure the gRPC `MaxReceiveMsgSize` and `MaxSendMsgSize` settings via environment variables and aims to help with #1331.

### Motivation
Per PR #2101 the gRPC settings for `MaxReceiveMsgSize` were set to 32 MB and for `MaxSendMsgSize` to 100 MB without the ability to overwrite them. In case of `MaxSendMsgSize` the value was significantly reduced from the [default](https://github.com/grafana/grafana-plugin-sdk-go/blob/8d134dc06e207d5b85d93a8edf6ad69921f32e61/backend/serve.go#L35) `math.MaxInt32` (approx. 2 GB). Depending on the Zabbix environment, Users may encounter gRPC message size limits when querying large datasets. This change allows administrators to adjust these values based on their specific needs.

### Changes
Added environment variable support for configuring gRPC message sizes by keeping the existing hard defaults as the fallback value. As far as I know, we can't use other plugin configuration options as the gRPC settings are directly configured on startup before the data source is initialized.

The environment variable names follow [Grafana's standard pattern](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#override-configuration-with-environment-variables) for plugin configuration: `GF_PLUGIN_<PLUGIN_ID>_<SETTING_NAME>`

#### Docker
```bash
docker run -d \
  -e "GF_PLUGIN_ALEXANDERZOBNIN_ZABBIX_DATASOURCE_GRPC_MAX_RECEIVE_MSG_SIZE_MB=64" \
  -e "GF_PLUGIN_ALEXANDERZOBNIN_ZABBIX_DATASOURCE_GRPC_MAX_SEND_MSG_SIZE_MB=200" \
  grafana/grafana
```

#### Docker Compose
```yaml
environment:
  - GF_PLUGIN_ALEXANDERZOBNIN_ZABBIX_DATASOURCE_GRPC_MAX_RECEIVE_MSG_SIZE_MB=64
  - GF_PLUGIN_ALEXANDERZOBNIN_ZABBIX_DATASOURCE_GRPC_MAX_SEND_MSG_SIZE_MB=200
```

### Backward Compatibility
This change is fully backward compatible. If the environment variables are not set, the plugin uses the same default values as before (32MB receive, 100MB send).


mario-zelger:
@zoltanbedi Maybe you can have a look at this?

mario-zelger:
@jcolladokuri Would it be possible to have a look at this? Thank you very much.

christos-diamantis:
Can we please take look at this? It affects a lot of our customers big environments

adamyeats (changes requested):
@mario-zelger Thanks for taking the time to put this together. Making these limits configurable is a reasonable ask. Unfortunately we can't take this in its current form, as configuration via environment variables doesn't translate to Grafana Cloud: customers there can't set env vars on their hosted instances, and the limits from #2101 are there to bound plugin memory usage on shared infrastructure.

There's also a practical snag on recent self-hosted Grafana: since grafana/grafana#77858, host environment variables are no longer forwarded to plugin subprocesses unless the plugin is opted in via `[plugins] forward_host_env_vars`, so these variables wouldn't actually reach the plugin's `os.Getenv` in a default setup.

The supported channel for this kind of config is the `[plugin.<plugin-id>]` ini section.  Grafana forwards those options to the plugin as `GF_PLUGIN_<KEY>` env vars, and that route stays compatible with Grafana Cloud. It would also be worth clamping the values to a sensible upper bound so the ~2GB default can't sneak back in.

If you're up for reshaping the PR along those lines, we'd be happy to take another look. I'll leave it open for now, no pressure either way.

mario-zelger:
@adamyeats Thanks for having a look at this. I'll try to see if I can reshape this so it fits the newer configuration model.

mario-zelger:
Need some more time but I'll still have a look at this.

christos-diamantis:
> Need some more time but I'll still have a look at this.

I can help you with reshaping this if you wish.

mario-zelger:
@christos-diamantis If you currently have time, feel free to to open a new PR based on this one or start fresh, if the codebase already changed too much. I'd love to see that feature rather sooner than later and I'm currently quite occupied with other projects.

If you start a new PR we can close this one afterwards. 👍

christos-diamantis:
Hello @mario-zelger. Just opened #2557 with the changes @adamyeats suggested.

PR #2243, opened 2026-01-23
feat: add support for filtering problems by cause/symptom

### LINKED ISSUE #2242: Add Support for Filtering Problems by Cause/Symptom
## Description
Zabbix 6.0+ introduced the ability to distinguish between cause and symptom problems through problem correlation. The Zabbix API's `problem.get` and `event.get` methods support a `symptom` parameter to filter problems accordingly, but the grafana-zabbix plugin currently does not expose this functionality in the UI.

## Feature Request
Add a new filter option in the Problems query editor to allow users to filter problems by their cause/symptom status.

### Proposed Solution
Add a new dropdown filter "Problem Type" in the Problems query editor with the following options:
- **All Problems** (default) - Shows both cause and symptom problems
- **Cause Only** - Shows only root cause problems (`symptom: false`)
- **Symptom Only** - Shows only symptom problems (`symptom: true`)

### Use Case
This feature is valuable for users who:
- Want to focus on root cause problems and ignore correlated symptoms
- Need to analyze symptom patterns separately from their causes
- Use Zabbix's problem correlation features extensively

### Expected Behavior
When a user selects:
- "All Problems" - No `symptom` parameter sent to API (default behavior)
- "Cause Only" - `symptom: false` parameter sent to `problem.get`/`event.get`
- "Symptom Only" - `symptom: true` parameter sent to `problem.get`/`event.get`

The filter should work for all problem query modes:
- Problems (current problems)
- Recent problems
- History (historical events)

christos-diamantis:
Closes #2242

## Description
This PR adds support for filtering Zabbix problems by their cause/symptom status, utilizing the `symptom` parameter available in Zabbix API's `problem.get` and `event.get` methods.

Zabbix 6.0+ introduced problem correlation features that distinguish between cause and symptom problems. This enhancement exposes that functionality in the grafana-zabbix plugin UI, allowing users to filter problems accordingly.

### DIFF
--- a/src/datasource/components/QueryEditor/ProblemsQueryEditor.tsx
+++ b/src/datasource/components/QueryEditor/ProblemsQueryEditor.tsx
@@ -31,6 +31,12 @@ const evaltypeOptions: Array<ComboboxOption<ZabbixTagEvalType>> = [
   { label: 'OR', value: ZabbixTagEvalType.Or },
 ];
 
+const symptomOptions: Array<ComboboxOption<string>> = [
+  { label: 'All Problems', value: 'all' },
+  { label: 'Cause only', value: 'false' },
+  { label: 'Symptoms only', value: 'true' },
+];
+
 export interface Props {
   query: ZabbixMetricsQuery;
   datasource: ZabbixDatasource;
@@ -244,6 +250,36 @@ export const ProblemsQueryEditor = ({ query, datasource, onChange }: Props) => {
             value={query.options?.severities}
             options={severityOptions}
             onChange={onSeveritiesChange}
+          />
+        </InlineField>
+        <InlineField label="Problem Type" labelWidth={16}>
+          <Combobox
+            width={24}
+            value={
+              query.options?.symptom === null || query.options?.symptom === undefined
+                ? 'all'
+                : query.options.symptom
+                ? 'true'
+                : 'false'
+            }
+            options={symptomOptions}
+            onChange={(option) => {
+              if (option) {
+                let symptomValue: boolean | null = null;
+                if (option.value === 'true') {
+                  symptomValue = true;
+                } else if (option.value === 'false') {
+                  symptomValue = false;
+                }
+                onChange({
+                  ...query,
+                  options: {
+                    ...query.options,
+                    symptom: symptomValue,
+                  },
+                });
+              }
+            }}
           />
         </InlineField>
       </QueryEditorRow>
--- a/src/datasource/components/QueryEditor/ProblemsQueryEditor.tsx
+++ b/src/datasource/components/QueryEditor/ProblemsQueryEditor.tsx
@@ -161,6 +161,7 @@ export const ProblemsQueryEditor = ({ query, datasource, onChange }: Props) => {
   };
 
   const supportsApplications = datasource.zabbix.supportsApplications();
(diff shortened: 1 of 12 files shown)


skaratsolis:
+1

src/datasource/components/QueryEditor/ProblemsQueryEditor.tsx
```diff
@@ -246,6 +252,36 @@ export const ProblemsQueryEditor = ({ query, datasource, onChange }: Props) => {
             onChange={onSeveritiesChange}
           />
         </InlineField>
+        <InlineField label="Problem Type" labelWidth={16}>
```
itsgareth: Cause / symptom problems were introduced in Zabbix 6.4 (https://www.zabbix.com/documentation/6.4/en/manual/introduction/whatsnew640), so i think symptom should be gated to 6.4+.

src/datasource/datasource.ts
```diff
@@ -611,6 +611,10 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery,
       problemsOptions.severities = severities;
     }
 
+    if (target.options?.symptom !== undefined && target.options?.symptom !== null) {
```
itsgareth, line 615: can we also add some tests for this feature

itsgareth (changes requested):
Thanks for the contribution! This is a clean additions, ive just left a few comments.

src/datasource/components/QueryEditor/ProblemsQueryEditor.tsx
christos-diamantis: Nice catch! I added a `supportsCauseSymptomProblems()` helper function on latest commit

src/datasource/datasource.ts
christos-diamantis: added tests for this.

itsgareth (dismissed):
LGTM, thank you for the contribution, @christos-diamantis!

itsgareth:
@christos-diamantis I can see your latest commits are signed, but the first few commits in this PR aren't, so the signature check is still failing. All commits in the PR need to be signed for it to pass.

Since it looks like you already have GPG signing set up, you can re-sign every commit in the branch with a rebase:

```
git rebase --exec 'git commit --amend --no-edit -n -S' main
git push --force-with-lease
```


This should walk through each commit and re-signs it with your key. Once all commits show as Verified and CI is green, we'll be able to merge.

christos-diamantis:
> @christos-diamantis I can see your latest commits are signed, but the first few commits in this PR aren't, so the signature check is still failing. All commits in the PR need to be signed for it to pass.
> 
> Since it looks like you already have GPG signing set up, you can re-sign every commit in the branch with a rebase:
> 
> ```
> git rebase --exec 'git commit --amend --no-edit -n -S' main
> git push --force-with-lease
> ```
> 
> This should walk through each commit and re-signs it with your key. Once all commits show as Verified and CI is green, we'll be able to merge.

I re-pushed. you can check if it is ok to merge :)

itsgareth (approved):

itsgareth:
@christos-diamantis thank you for your contribution here, its very much appreciated! ill speak with the team and see when we next plan to create a release for this data source.

PR #2256, opened 2026-01-26
feat: create new multi-metric table query type

### LINKED ISSUE #2255: Feature Request: Multi-metric table query type
## Summary
Add a new query type "Multi-Metric Table" that allows creating dynamic tables where rows represent entities (e.g., network interfaces, disks, filesystes, containers etc) and columns represent different metrics for those entities. This is really useful for creating tables for multiple similar items accross one or more hosts. 

Currently, to display metrics for multiple similar items in tabular format, a user needs to:
- Create queries selecting a lot of items
- The plugin retrieves timeseries for all the items for the timespan selected
- Heavy processing needed afterwards on frontend on "Calculations" section of the panel (Reduce, Group by, Extract fields etc)

The current approach is time-consuming, error-prone and does not scale well for displaying data for big number of entities.

christos-diamantis:
This PR closes #2255 

## Summary
Adds a new **Multi-Metric Table** query type: define an entity pattern that becomes the table rows (with optional multi-capture-group extraction into columns), then add metric columns — each with its own pattern, aggregation, and an optional sparkline. Entities are discovered automatically, and the default `last` aggregation reads Zabbix `lastvalue` in bulk instead of fetching history, which keeps large tables fast.

## Features
**Dynamic entity rows** — pattern-based item selection (by name or key); a regex with multiple capture groups decomposes complex item names into separate columns; entities are de-duplicated automatically.
- Example: `Interface eth0[WAN]: Bits received` with extract `Interface (.*)\[(.*)\]: .*` → column 1 = `eth0`, column 2 = `WAN`.

**Metric columns** — add any number; each has its own pattern, search-by name/key, and aggregation:
- `last` (default) — uses the item's `lastvalue`, no history fetched (fast, ideal for many entities).
- `avg`, `min`, `max`, `sum`, `median`, `p95` — fetch history for the panel time range and aggregate.

**Live suggestions** — group, host, host-tag and all pattern fields are autocomplete dropdowns populated from the datasource (same UX as the other query editors); custom regex / `/.../` values are still allowed.

**Sparkline columns (optional, default off)** — a per-column toggle that returns the column's full history as its own time-series frame. Render it by adding Grafana's **Time series to table** transformation (one Trend column per sparkline metric), then **Join by field** / **Merge** on the Host/Entity columns. Series are grid-aligned so they draw as continuous lines.

## Screenshots
**Query editor**

*Selecting group and host:*

Screenshot 2026-06-16 at 10 16 06

*Defining pattern for metadata columns (pattern extraction) - Metric picker auto completes as other panels:*

Screenshot 2026-06-16 at 10 16 18

Screenshot 2026-06-16 at 10 16 26

*Defining metric columns - Metric picker auto completes as other panels:*

Screenshot 2026-06-16 at 10 16 47

Screenshot 2026-06-16 at 10 16 55

*Resulting table:*

Screenshot 2026-06-16 at 10 17 30

**Table with sparkline columns**

*Enabling the Sparkline on the metric columns, then applying 2 transformations (1 - Time series to table, 2 - Merge series/tables):*

Screenshot 2026-06-16 at 10 17 40

Screenshot 2026-06-16 at 10 17 58

*Resulting table:*

Screenshot 2026-06-16 at 10 18 10

## Performance / benchmark
The `last` aggregation reads `lastvalue` in two bulk `item.get` calls instead of pulling full history, so cost scales with entity count rather than time range. A benchmark reproduces the comparison against the plain one-query-per-column approach (counts mirror the real backend call graph):

```bash
go run ./bench/multimetric                                   # defaults: 10 hosts × 8 interfaces, 1h
go run ./bench/multimetric -hosts 100 -ifaces 24 -range 24h  # large dashboard
go run ./bench/multimetric -metrics 2 -multi-non-last 2      # both columns use a non-"last" aggregation
```

Representative results (`last` aggregation, 2 columns, 50ms/RPC) — full details in [`bench/multimetric/README.md`](bench/multimetric/README.md):

| Scenario | Plain (one query/column) | Multi-metric table |
| --- | --- | --- |
| 10 hosts × 8 ifaces, 1h | 1.33 MiB | 753 KiB (**1.8× less**) |
| 50 hosts × 16 ifaces, 6h | 46.9 MiB | 3.67 MiB (**12.8× less**) |
| 100 hosts × 24 ifaces, 24h | 533 MiB | 7.34 MiB (**72.6× less**) |

The payload win grows with time range × interface count; it narrows for columns using a non-`last` aggregation or a sparkline, since those fetch history again.


skaratsolis:
+1

christos-diamantis:
Updated the PR body. 
I refurbished the inital PR code with help of Claude and added the auto-suggestion dropdown menu on all comboboxes.
Also, added the feature for displaying metrics as sparkline on the resulting table. Added screenshots on the PR description in order to quickly understand how it works.

Finally, created benchmark for true comparison of Zabbix API calls when trying to build a table using the `Metrics` query type and then applying heavy transformations, and my way of batch getting the last value of all items.

Need to understand that this PR offers huge performance win for displaying a table with last values as well as provides flexibility for creating complex tables without the need of complex transformations in Grafana.

@zoltanbedi please for your review on this. It's been a long time its open

hbertorello-ggta:
Hi! I am heavily testing a custom workaround for this exact issue in my production environment because creating inventory tables in Grafana with Zabbix is currently a major pain point. Relying on complex transformations like 'Reduce' and 'Merge' severely impacts browser performance when dealing with a high volume of hosts.

Having a native way to output a flat, wide dataframe grouped by Host for both Metrics and Text queries is absolutely essential for asset management dashboards. I really hope this PR gets reviewed and merged soon. Thanks for the hard work!

temp

christos-diamantis:
ran some tests and fixed some issues i found:
1. i could not select a string/text/log as "metric column". added the option for it. when selecting `Text` as Value type, `Last` aggregation is only supported and Sparkline is obviously disabled.
2. even though a "Global" pattern was set on the `Entity Pattern` section, the `Pattern` on the metric column section was not a subset of the global pattern. you could select anything which was not correct.
3. The whole idea behind `Multi-metric table` is focused primarily on optimal tabular display of LLD items from Zabbix, but should not only be. There are valid cases of showing a table with multiple different metrics for multiple hosts which are not LLD items. e.g. Current CPU and memory utilization for X number of hosts. In order for this to be feasible a `Rows from` switch has been added. `Hosts` selection will provide the ability to do so.
4. Some retouch on the tooltips, placeholders and the general help text - when Multi-metric table is selected - has been done for it to become more user friendly.

Waiting for review from someone + any possible feedback is really welcome :)

hbertorello-ggta:
I have been struggling with this exact issue and using direct DB queries. This feature would be incredibly helpful! Looking forward to testing it.

christos-diamantis:
Any update?

Issue #2266, opened 2026-01-28
TypeError in Service Query when property is not 'status'


petersonbasso:
**Describe the bug**
A TypeError: Cannot read properties of null (reading '1') occurs in the Grafana Zabbix plugin when using a Service query type and selecting any property other than status (e.g., sli, uptime, downtime).

**Expected behavior**
The query should execute normally using the default or specified interval, without throwing a TypeError.

**Steps to reproduce**
1. Create a new dashboard or panel.
2. Select a Zabbix Datasource.
3. Change the Query Mode to Services.
4. Select any Service and SLA.
5. Change the Property from status to sli (or any other property except status).
6. Observe the console error or the panel failing to load.

**Actual behavior**
The plugin crashes with the following error in the browser console:
```
runRequest.ts:166 runRequest.catchError TypeError: Cannot read properties of null (reading '1')
    at ge (utils.ts:264:43)
    at Et.<anonymous> (zabbixAPIConnector.ts:491:24)
    at Generator.next (<anonymous>)
    at gt (module.js?_cache=6.1.1:4:58366)
    at a (module.js?_cache=6.1.1:4:58569)
    at module.js?_cache=6.1.1:4:58630
    at new Promise (<anonymous>)
    at Et.<anonymous> (module.js?_cache=6.1.1:4:58510)
    at Et.getSLI (zabbixAPIConnector.ts:505:3)
    at Vt.<anonymous> (zabbix.ts:692:43)
```

**Software versions**

| Grafana | Zabbix | Grafana-Zabbix Plugin |
| ------- | ------ | --------------------- |
| 12.3.1   | 7.0.22  | 6.1.2                 |

PR #2274, opened 2026-01-29
Fix service query when selecting an SLA

jcolladokuri:
## Summary

As part of the migration to `DatasourceWithBackend` in #2123 we failed to introduce  as part of the request the `slaInterval` this resulted in an error, because it is used in a calculation for an `intervals` parameter before we send the request to the backend. This PR reads the `slaInterval` when querying for services

## Details

1. Readds `slaInterval` to pass down before querying the backend
2. Instead of adding it as part of the request, `slaInterval` is now passed down as a function argument for the SLI and SLA functions. 

    a. Why? Previously we were able to set the `slaInterval` on the request because we did not enforce any types on that argument. However, now that we have a stricter type (namely `DataQueryRequest<ZabbixMetricsQuery>`) we are not able to set a nonexistent property on `request`. Instead we read it form the target, and pass it down into the respective SLA and SLI functions.

Fixes: #2266


zoltanbedi (approved):
Thank you!

PR #2282, opened 2026-02-04
Replace use of `api/datasources` with `api/datasources/uid`

### LINKED ISSUE #2262: Migrate data source id to datasource uid for direct db connections
Migrate data source id to datasource uid for direct db connections

jcolladokuri:
## Summary

The `api/datasources` API will no longer be supported. This PR replaces all uses of it with `api/datasources/uid` throughout.

## Detailed summary

1. Changes the API URL itself to be `api/datasources/uid/<datasource_uid>`
2. Replaces all uses of datasource ID with datasource UID for direct DB connections:
    a. Changes the value of dropdown options in the config editor to be `uid` instead of `id`
    b. Finds datasources by `uid` using `DatasourceSrv` instead of `id`
    c. Stores the datasource `uid` in `jsonData` instead of datasource `id`
    d. For older datasource JSON schemas where `dbConnection` is used, it uses the datasource ID to find the `uid` and use that throughout
    e. It stores the datasource type so that for SQL queries, we run the query with the proper datasource object 
3. Overall, adds stronger typing for anything related to `jsonData`, `dbConnections`, `datasourceUIDs`
4. Adds tests.

How to test
1. With DB connections:
    a. Create a zabbix datasource and enable Direct DB connection
    b. Select a sql, or influx db datasource
    c. Run queries for all different query types
3. Without DB connections:
    a. Create a zabbix datasource, click save & test
    b. Run queries for all different query types (`Metrics`, `Text`, `Services` etc)
3. Enable `datasourceDisableIdApi` locally, and run all queries, save & test as usual, you should not run into any problems

## Known issues
While testing the changes for this PR:

- I discovered a bug with influx db direct connections. See open issue [here](https://github.com/grafana/grafana-zabbix/issues/2281)

Fixes #2262

### DIFF
--- a/src/datasource/migrations.ts
+++ b/src/datasource/migrations.ts
@@ -162,51 +162,39 @@ function convertToRegex(str) {
 }
 
 export function migrateDSConfig(jsonData: ZabbixDSOptions) {
-  if (!shouldMigrateDSConfig(jsonData)) {
+  if (!jsonData) {
     return jsonData;
   }
 
-  const oldVersion = jsonData.schema || 1;
-
-  if (oldVersion < 2 && jsonData.dbConnection) {
+  // Migrate nested dbConnection object (schema v1) to flat fields
+  if (jsonData.dbConnection) {
     const dbConnectionOptions = jsonData.dbConnection;
-    jsonData.dbConnectionEnable = dbConnectionOptions.enable;
-    jsonData.dbConnectionDatasourceId = dbConnectionOptions.datasourceId;
+    jsonData.dbConnectionEnable = dbConnectionOptions.enable || false;
+    if (!jsonData.dbConnectionDatasourceUID && dbConnectionOptions.datasourceId > 0) {
+      jsonData.dbConnectionDatasourceUID = getUIDFromID(dbConnectionOptions.datasourceId);
+    }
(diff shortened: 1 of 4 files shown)


src/datasource/migrations.ts
```diff
@@ -172,20 +170,31 @@ export function migrateDSConfig(jsonData) {
   jsonData.schema = DS_CONFIG_SCHEMA;
 
   if (oldVersion < 2) {
-    const dbConnectionOptions = jsonData.dbConnection || {};
-    jsonData.dbConnectionEnable = dbConnectionOptions.enable || false;
-    jsonData.dbConnectionDatasourceId = dbConnectionOptions.datasourceId || null;
+    // disabling as it is currently needed for migration
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
+    const dbConnectionOptions = jsonData.dbConnection;
+    jsonData.dbConnectionEnable = dbConnectionOptions?.enable || false;
```
yesoreyeram: @jcolladokuri - Can you help me understand why this migration is under the condition `oldVersion < 2`

yesoreyeram (changes requested):
Overall LGTM. But, the datasources provisoned with old properties failing. Need to fix

PR #2291, opened 2026-02-10
Chore: Batch dependency update

itsmylife:
Fixes: https://github.com/grafana/oss-big-tent-squad/issues/174

### DIFF
--- a/.changeset/neat-emus-listen.md
+++ b/.changeset/neat-emus-listen.md
@@ -1,0 +1,5 @@
+---
+'grafana-zabbix': minor
+---
+
+Update dependencies for react-19 upgrade preparations


.github/workflows/publish.yaml
```diff
@@ -30,11 +30,9 @@ jobs:
...
     with:
-      golangci-lint-version: '2.7.2'
-      go-version: '1.25.5'
       branch: ${{ github.event.inputs.branch }}
       environment: ${{ github.event.inputs.environment }}
       docs-only: ${{ fromJSON(github.event.inputs.docs-only) }}
       run-playwright: true
       github-draft-release: false
-      run-playwright-with-skip-grafana-react-19-preview-image: true
+      run-playwright-with-skip-grafana-react-19-preview-image: false
```
zoltanbedi: We can just remove this as it is false by default.

PR #2416, opened 2026-05-12
feat: add column sorting and free-text search filter on problems panel

### LINKED ISSUE #2415: Problems panel - Add dynamic column sorting + global search
The Zabbix Problems panel currently has no way to sort columns or filter the displayed problems client-side. 
Pre migrating to React Table v8, sorting was possible but now it is not. Users with large problem lists have no way to quickly find specific problems or reorder them without changing the datasource query.

### Proposed Behaviour

**Column sorting:**
- Clicking a column header sorts the table by that column (ascending)
- Clicking again toggles to descending
- A sort indicator (▲ / ▼) is shown on the active column
- Sortable columns: **Host**, **Host (Technical Name)**, **Proxy**, **Severity**, **Problem**, **Datasource**, **Age**, **Time**, **Custom tag columns**
- Default sort: **Time descending** (most recent first)

**Free-text search filter:**
- A search input appears above the table and filters rows across all visible columns in real time
- The filter is **disabled by default** and can be enabled per-panel via a new panel option: _"Show search filter"_
- Resetting the input restores all rows
- Changing the search resets pagination back to page 1

### Implementation Notes
The panel already uses **React Table v8**, which provides `getSortedRowModel` and `getFilteredRowModel` out of the box. No external dependencies are needed — this is purely enabling built-in functionality.

christos-diamantis:
This PR closes #2415 

## Summary

Adds column sorting and an optional free-text search filter to the Zabbix Problems panel.

## Features

**Column sorting:**
- Clicking a column header sorts the table by that column (ascending)
- Clicking again toggles to descending
- A sort indicator (▲ / ▼) is shown on the active column
- Sortable columns: **Host**, **Host (Technical Name)**, **Proxy**, **Severity**, **Problem**, **Datasource**, **Age**, **Time**, **Custom tag columns**
- Default sort: **Time descending** (most recent first)

**Free-text search filter:**
- A search input appears above the table and filters rows across all visible columns in real time
- The filter is **disabled by default** and can be enabled per-panel via a new panel option: _"Show search filter"_
- Resetting the input restores all rows
- Changing the search resets pagination back to page 1

## Screenshots

Screenshot 2026-05-12 at 14 50 00
Screenshot 2026-05-12 at 14 50 50


skaratsolis:
+1

G0nz0uk:
This would be an amazing addition,  any idea when this might get approved and added?

christos-diamantis:
@zoltanbedi any update? I do not see any reviewer assigned unfortunately.

src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -264,6 +285,19 @@ export const ProblemList = (props: ProblemListProps) => {
   );
   const [columnResizeMode] = useState<ColumnResizeMode>('onChange');
 
+  const [sorting, setSorting] = useState<SortingState>([
```
akkikumar72: **The hardcoded initial sort silently overrides the existing "Sort by" panel option.** `ProblemsPanel` already orders rows through `sortTriggers` from `panelOptions.sortProblems` (Default / Last change / Severity). Once `getSortedRowModel` is registered with a fixed `lastchange desc` state, that ordering is re-applied on top and the configured one is lost.
I built the base branch and this branch against the same six seeded problems, with the panel set to `sortProblems: 'priority'`:
```
base:  Disaster, High, High, Average, Average, Information   (severity desc, as configured)
this:  Disaster, Average, High, Information, Average, High   (time desc, option ignored)
```
Every existing dashboard on "Severity" or "Default" changes ordering after upgrade, and the only way back is clicking a header (which does not persist). Deriving the initial state from the option keeps them working:
```suggestion
  const [sorting, setSorting] = useState<SortingState>(() => {
    switch (panelOptions.sortProblems) {
      case 'priority':
        return [{ id: 'priority', desc: true }];
      case 'lastchange':
        return [{ id: 'lastchange', desc: true }];
      default:
        return [];
    }
  });
```

src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -413,13 +455,39 @@ export const ProblemList = (props: ProblemListProps) => {
...
           <thead>
             {table.getHeaderGroups().map((headerGroup) => (
               <tr key={headerGroup.id}>
                 {headerGroup.headers.map((header) => (
-                  <th key={header.id} style={{ width: `${header.getSize()}px` }}>
-                    {header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
+                  <th
+                    key={header.id}
+                    style={{ width: `${header.getSize()}px` }}
+                    onClick={header.column.getCanSort() ? header.column.getToggleSortingHandler() : undefined}
```
akkikumar72: **Resizing a column also re-sorts the table.** The click handler is on the whole `th` and the resizer is a child of it. TanStack's resize handler only takes `mousedown`/`touchstart` and does not stop propagation, so the browser still dispatches a `click` on the `th` when a drag starts and ends inside the header cell.
Clicking directly on `div.resizer` in the running panel cleared the active `Host ▲` sort and set `Severity ▼`, with the column width unchanged at 104px, so it is a pure side effect:
```
before: Host ▲ | Severity ⇅   severityWidth=104
after:  Host ⇅ | Severity ▼   severityWidth=104
```
Moving the handler onto the header content span keeps sorting and resizing separate:
```suggestion
                  <th
                    key={header.id}
                    style={{ width: `${header.getSize()}px` }}
                    className={header.column.getCanSort() ? 'sortable-header' : ''}
                  >
                    <span
                      className="header-content"
                      onClick={header.column.getCanSort() ? header.column.getToggleSortingHandler() : undefined}
                    >
```

src/panel-triggers/components/Problems/Problems.tsx
akkikumar72: `yarn lint` fails on this line, which will block CI:
```
288:57  error  Replace `⏎····{·id:·'lastchange',·desc:·true},⏎··` with `{·id:·'lastchange',·desc:·true·}`  prettier/prettier
```
`src/styles/_panel-problems.scss` also fails `prettier --check` (no trailing newline). `yarn lint:fix` clears both.

akkikumar72 (commented):
Thanks for working on it. Added some comments

src/panel-triggers/components/Problems/Problems.tsx
christos-diamantis: Thank you for the suggestion @akkikumar72. Indeed that would break already existing dashboards.
I changed the code based on your suggestion and also added a useEffect() so when the panel options are updated, the table is updated also showing the correct sorting selected.
linting issues fixed also

src/panel-triggers/components/Problems/Problems.tsx
christos-diamantis: pretty nice find right here. i never found it on my own tests. fixed based on your suggestion

christos-diamantis:
@akkikumar72 Thank you for the review.
The problems are fixed: 
- The sorting now reads the pre-selected panel options for sorting by and repects this, so no already existing dashboard is affected
- Added an additional `useEffect` block so that changing the panel Options reflects immediately on the table
- The handler has been moved out of `th` to the `header-content` so the resize does not trigger sorting
- linting issues fixed
- fixed tests for Age, Severity and Datasource columns which were failing. The test tried to do a literal match with the header but now it has the sorting svg, so changed it to `includes()`

src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -208,6 +239,7 @@ export const ProblemList = (props: ProblemListProps) => {
         id: 'age',
         header: 'Age',
         size: 100,
+        enableSorting: true,
```
akkikumar72, line 257: `Age` sorts in the opposite direction to its indicator. The accessor is `timestamp` and there's no comparator, so ascending order puts the *oldest* problem first — the column reads "3 days, 12 hours, 5 minutes" under a `▲`.

Reproduced against the same three rows:

```
Age ASC (header shows "▲")  ->  72h, 12h, 1h
```

Age counts backwards from `timestamp`, so the comparator needs inverting:

```suggestion
        enableSorting: true,
        // Age counts backwards from timestamp: a newer timestamp is a smaller age.
        sortingFn: (rowA, rowB) => Number(rowB.original.timestamp) - Number(rowA.original.timestamp),
```

With that applied: `Age ASC -> 1h, 12h, 72h` and `Age DESC -> 72h, 12h, 1h`.

src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -194,6 +224,7 @@ export const ProblemList = (props: ProblemListProps) => {
       columnHelper.accessor('datasource', {
         header: 'Datasource',
         size: 120,
+        enableSorting: true,
```
akkikumar72, line 246: This column has two problems once sorting is enabled, because `ProblemDTO.datasource` is `DataSourceRef | string`:

- when it holds a `DataSourceRef` object, it hits the same `compareBasic` / `[object Object]` fallback as `Host Groups`, so clicking just reverses rows
- when it holds a string, it sorts the raw uid while the cell renders the *resolved* instance name from `getDataSourceSrv()`, so the order won't match the visible text

Extracting the name resolution the cell already does makes both consistent:

```ts
// module level, next to buildCustomTagColumns
const resolveDatasourceName = (datasource?: DataSourceRef | string): string => {
  if ((datasource as DataSourceRef)?.uid) {
    const instance = getDataSourceSrv().getInstanceSettings((datasource as DataSourceRef).uid);
    return instance?.name ?? String((datasource as DataSourceRef).uid);
  }
  return (datasource as string) ?? '';
};
```

```ts
sortingFn: (rowA, rowB) =>
  resolveDatasourceName(rowA.original.datasource).localeCompare(resolveDatasourceName(rowB.original.datasource)),
cell: ({ cell }) => <span>{resolveDatasourceName(cell.getValue())}</span>,
```

src/panel-triggers/components/Problems/Problems.tsx
```diff
@@ -106,11 +123,13 @@ export const ProblemList = (props: ProblemListProps) => {
       columnHelper.accessor('host', {
         header: 'Host',
         size: 120,
+        enableSorting: true,
```
akkikumar72, line 141: The string columns rely on the default `sortingFn: 'auto'`, and that makes their ordering depend on how many problems are in the table. `getAutoSortingFn` samples `flatRows.slice(10)` — rows 11 onward, not the first 10 — so a table with 10 or fewer rows never sees a string value and falls through to `basic` (case-sensitive lexicographic), while 11+ rows resolves to `alphanumeric` (natural):

```
10 rows -> comparator=basic         Host ASC -> SRV3, srv1, srv10, srv2
11 rows -> comparator=alphanumeric  Host ASC -> srv1, srv2, SRV3, srv10
```

Same four hosts, same click, different order — purely because a problem was added elsewhere in the table. Pinning the comparator explicitly makes it stable at any row count:

```ts
sortingFn: 'alphanumeric',
```

Applies to `host` (L126), `hostTechName` (L132), `proxy` (L146), `name` (L195), `opdata` (L198, sortable by default), and the columns built in `buildCustomTagColumns`. Verified `srv1, srv2, SRV3, srv10` at both 4 and 24 rows afterwards.

akkikumar72:
Thanks for turning these around so quickly, I re-checked both fixes and they're fine.

I ran the checks locally against `493d8c5`.

One request: could you add a few tests for the new behaviour? Right now the only test change relaxes three existing header assertions to `includes()`, so nothing covers sorting or search. Three small cases in the existing `Problems.test.tsx` would do it, and two of them would have caught the `Age` direction and `Host Groups` findings above on their own:

- clicking a sortable header reorders rows
- typing in the search input filters rows
- a given `sortProblems` option produces the expected initial order

Also worth noting the remaining `prettier --check` failure on `_panel-problems.scss` is entirely pre-existing lines (`content: ""` quoting, a long `linear-gradient`, a multi-value `box-shadow`), please don't reformat those here, the trailing-newline fix was the right scope.

src/panel-triggers/components/Problems/Problems.tsx
christos-diamantis: applied your suggested fix. indeed the sorting was reverse than the expected one

src/panel-triggers/components/Problems/Problems.tsx
christos-diamantis: applied the change to datasource column as well as to the groups one where the same issue was existing

src/panel-triggers/components/Problems/Problems.tsx
christos-diamantis: pinned `sortingFn` to `alphanumeric` as suggested cause that could be a real issue on such cases.

christos-diamantis:
@akkikumar72 I fixed all the sorting anomalies you found and created some tests covering all what you found on this review as well as covering the functionality in a whole. Specifically, added tests:

- clicking a sortable header reorders rows and toggles the indicator
- the sortProblems panel option produces the expected initial order wihtout any header click
- typing in the search input filters rows case-insesitively across columns and clearing it restores all rows
- the sort indicator is rendered on all 11 sortable columns and is absent on Status, Status Icon, Ack and Tags
- every sortable columns sorts correctly in both directions

Thank you for your time reviewing this. Appreciate it.

akkikumar72 (approved):
Approving now. 
Nice work @christos-diamantis , and thanks for your patience. We will include this fix in next release

PR #2423, opened 2026-05-19
Host tag filter in template variables + inherited tag support

### LINKED ISSUE #1682: Host filter by tag value (#1651)
Hello! Please **@alexanderzobnin**, as commented by **@xocoru** at https://github.com/grafana/grafana-zabbix/issues/1651

It would be great to be able to **filter hosts by name and/or tag value.**

Some organizations have tags that represent business units, but we cannot change hostnames or create multiple host groups and filtering by host tag would solve this. The idea is to add the feature and not change the current context which is already fantastic!

image

Thank you very much!

MaxDiOrio:
## Summary

- Add a **Host tag** picker to the dashboard variable query editor (`Host`, `Application`, `ItemTag`, `Item`, `ItemValues` types). Closes #1682. Variable support was deferred from PR #2140 and the carryover comment from @mcbrineellis went unanswered, so this picks it up.
- Extend host-tag matching to **template-inherited tags** in both the panel editor and the new variable editor — previously the server-side `tags` filter on `host.get` silently dropped hosts whose match came from a linked template.
- Add **tag-name and tag-value autocomplete** in the host tag picker, populated from the merged direct + inherited tag inventory.
- Restore saved host-tag filters when reopening a dashboard via a new `value` prop on `HostTagQueryEditor` (the editor previously always started empty).

## Why

Existing template variable types (Group / Host) only filter by host name or group, forcing users to either rename hosts to encode dimensions or create one host group per dimension. Tags are the natural place to encode `class:database`, `platform:globalspec`, etc., and PR #2140 wired this up for panel queries but not for template variables. This finishes the feature.

The inherited-tag change came out of dogfooding against a Zabbix 7.4 install where most tag values come from linked templates. Without `selectInheritedTags` + client-side evaluation, the filter returns zero hosts for those tags — which is unintuitive and was hitting users of the existing panel-editor feature too.

### Variable editor (new)
A `Host tag` row renders for all non-Group query types. It uses the same `HostTagQueryEditor` component as the panel editor — same operators (Contains, Equals, DoesNotContain, DoesNotEqual, Exists, DoesNotExist), same AND/OR toggle. When `hostTags` are set and the Host filter is empty, `metricFindQuery` coerces Host to `/.*/` so an empty Host means "all hosts matching the tags" (it would otherwise return `[]` due to `findByName('')` semantics).

### Inherited tag matching (panel + variable)
`zabbixAPIConnector.getHosts` now:
1. Requests `selectInheritedTags: 'extend'` alongside `selectTags: 'extend'` when tags are needed (5.4+).
2. Merges `tags` + `inheritedTags` into a single `tags` array, de-duped on `tag\tvalue`.
3. Drops the server-side `tags` / `evaltype` parameters (would only match direct tags) and applies the filter client-side against the merged list via a new `matchHostTag` helper that mirrors all six operators.

Performance tradeoff: we fetch the host list for the group without server-side narrowing, then filter in JS. For typical group sizes (dozens–hundreds of hosts) this is a non-issue. For very large groups, a future enhancement could combine the existing server-side direct-tag filter with a supplementary inherited-tag pass and union the results.

### Autocomplete
The tag-name Combobox already existed; this PR also builds a `{tagName → unique values}` map from the same fetched hosts and threads it through to the value Combobox, which now lists known values for the selected tag while still allowing custom free-text input (needed for Contains-style filters against substrings that don't appear in the inventory).

### `HostTagQueryEditor`
Adds an optional `value?: HostTagFilter[]` prop, used to seed `useState` so saved filters reappear on dashboard reload. Existing callsites that don't pass it keep the prior empty-on-mount behaviour.

## Test plan
- [x] `yarn typecheck` — clean
- [x] `yarn lint` — 0 errors (existing 65 deprecation warnings unchanged)
- [x] `yarn jest` — 177/177 pass, including 2 new tests in `zabbixAPIConnector.test.ts` covering client-side inherited-tag filtering and operator dispatch
- [x] Manual smoke test against a real Zabbix 7.4.9 datasource:
  - Variable editor: Host query type with `class Contains database` returns the database hosts (whose `class` tag is inherited from a Windows-server template, never directly assigned). Tag and value Comboboxes both autocomplete from the host inventory.
  - Panel editor (Explore): same filter shape, same result.
  - Multiple tag filters with AND/OR toggle, Exists / DoesNotExist operators, saved-state repopulation on reload — all behave correctly.

## Files changed (10)
- `CHANGELOG.md` — unreleased section
- `src/datasource/types.ts` — extend `VariableQuery` with `hostTags`, `evaltype`
- `src/datasource/components/VariableQueryEditor.tsx` — render `HostTagQueryEditor` for non-Group queries; new `HostTagFilterLoader` functional sub-component loads autocomplete options
- `src/datasource/components/QueryEditor/HostTagQueryEditor.tsx` — add `value` and `hostTagValueOptions` props; value field becomes a Combobox with `createCustomValue`
- `src/datasource/components/QueryEditor/MetricsQueryEditor.tsx` — extend `loadHostTagOptions` to return both tag-name and tag-value option maps
- `src/datasource/datasource.ts` — pass `hostTags`/`evaltype` through `metricFindQuery`; interpolate tag key/value
- `src/datasource/zabbix/zabbix.ts` — `getHosts` signature forwards tag filters to `getAllHosts`
- `src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts` — request `selectInheritedTags`, merge tag lists, apply client-side filter with new `matchHostTag` helper
- `src/datasource/specs/datasource.spec.ts` — update `getHosts` arity, add `hostTags`/`evaltype` propagation test
- `src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.test.ts` — adjust expectations for client-side filtering, add inherited-tag coverage

Happy to split the inherited-tag change into a separate PR if reviewers prefer, but the variable editor would feel half-finished without it on real-world Zabbix deployments using template-driven tagging.


src/datasource/datasource.ts
```diff
@@ -731,14 +731,29 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery,
...
+      value: utils.replaceTemplateVars(this.templateSrv, f.value, {}),
+    }));
 
     switch (queryModel.queryType) {
       case VariableQueryTypes.Group:
         resultPromise = this.zabbix.getGroups(queryModel.group);
         break;
-      case VariableQueryTypes.Host:
-        resultPromise = this.zabbix.getHosts(queryModel.group, queryModel.host);
+      case VariableQueryTypes.Host: {
```
adamyeats: The editor renders the Host tag row for every non-Group query type, but only the Host branch passes `interpolatedHostTags` through. A tag filter on an Application, Item tag, Item or Item values variable is saved and displayed but silently ignored, which the PR description claims is supported. Consider either wiring the filter through for those types or only rendering the row for Host queries.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
```diff
@@ -162,24 +163,60 @@ export class ZabbixAPIConnector {
...
+      const activeFilters = hostTagFilters.filter((f) => f.tag !== '');
+      if (activeFilters.length === 0) {
+        return merged;
+      }
+      const isOr = evalType === ZabbixTagEvalType.Or;
+      return merged.filter((h: any) => {
+        const tags = (h.tags ?? []) as Array<{ tag: string; value?: string }>;
+        return isOr
+          ? activeFilters.some((f) => matchHostTag(tags, f))
+          : activeFilters.every((f) => matchHostTag(tags, f));
```
adamyeats: Zabbix evaltype 0 is And/Or, which ORs filters sharing a tag name and ANDs across different names, but `every` is a strict AND. Two filters on the same tag (for example `role Equals api` plus `role Equals web`) previously matched hosts with either value and now require a host to carry both, a behaviour regression for the existing panel feature. Consider mirroring the And/Or grouping (OR within a tag name, AND across names) in the client-side matcher.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
```diff
@@ -162,24 +163,60 @@ export class ZabbixAPIConnector {
...
-      // convert operator to number since that is the expected type in Zabbix.
-      params.tags = hostTagFilters
-        .filter((tagFilter) => tagFilter.tag !== '')
-        .map((tagFilter) => {
-          return { ...tagFilter, operator: +tagFilter.operator };
-        });
+      // Also fetch tags inherited from linked templates so callers can see / filter on them.
+      // The host.get `tags` filter param only matches direct host tags — inherited matches must be
+      // applied client-side. selectInheritedTags is supported from Zabbix 5.4.
+      if (this.version && semver.gte(this.version, '5.4.0')) {
```
adamyeats: `this.version` is set asynchronously by `initVersion()`, so a `getHosts` call that races it builds params without `selectInheritedTags` and silently drops inherited matches, and `CachingProxy` then caches that wrong result for 10 minutes. Consider awaiting the version before building params, as `getAllApps` does with `getVersion()`.

src/datasource/datasource.ts
```diff
@@ -731,14 +731,29 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery,
 
     queryModel = queryModel as VariableQuery;
     const { group, host, application, item } = queryModel;
+    const interpolatedHostTags = queryModel.hostTags?.map((f) => ({
+      ...f,
+      tag: utils.replaceTemplateVars(this.templateSrv, f.tag, {}),
+      value: utils.replaceTemplateVars(this.templateSrv, f.value, {}),
```
adamyeats: `replaceTemplateVars` wraps any substituted value in `/^...$/`, but `matchHostTag` compares literally, so a tag value of `$env` interpolates to `/^production$/` and never matches. Consider using plain `templateSrv.replace` here (the same applies to `f.tag`), or teaching the matcher to honour regex values.

src/datasource/components/QueryEditor/MetricsQueryEditor.tsx
```diff
@@ -245,6 +270,7 @@ export const MetricsQueryEditor = ({ query, datasource, onChange, onItemCountCha
         <InlineField label="Host tag" labelWidth={12}>
           <HostTagQueryEditor
             hostTagOptions={hostTagsOptions}
+            hostTagValueOptions={hostTagValueOptions}
```
adamyeats, line 244: The panel editor does not pass the new `value` prop, so saved panel-query tag filters still start empty on reopen and the mount effect in `HostTagQueryEditor` then writes an empty `hostTags` back into the query. Consider passing `value={query.hostTags}` here too, otherwise the changelog claim only holds for variables.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
```diff
@@ -1044,6 +1081,32 @@ export class ZabbixAPIError {
...
+      return !tags.some(sameKey);
+    case HostTagOperatorValue.Equals:
+      return tags.some((t) => sameKey(t) && (t.value ?? '') === filter.value);
+    case HostTagOperatorValue.DoesNotEqual:
+      return tags.every((t) => !sameKey(t) || (t.value ?? '') !== filter.value);
+    case HostTagOperatorValue.Contains:
+      return tags.some((t) => sameKey(t) && (t.value ?? '').includes(filter.value));
+    case HostTagOperatorValue.DoesNotContain:
+      return tags.every((t) => !sameKey(t) || !(t.value ?? '').includes(filter.value));
+    default:
```
adamyeats, line 1216: An unrecognised operator silently behaves like Contains. Consider returning false instead, which avoids matching on a value the user never asked for.

src/datasource/components/QueryEditor/HostTagQueryEditor.tsx
```diff
@@ -17,12 +19,14 @@ export const HostTagQueryEditor = ({
...
   version,
+  value,
   evalTypeValue,
+  hostTagValueOptions,
   onHostTagFilterChange,
   onHostTagEvalTypeChange,
 }: Props) => {
-  const [hostTagFilters, setHostTagFilters] = useState<HostTagFilter[]>([]);
-  const [hostTagValueDrafts, setHostTagValueDrafts] = useState<string[]>([]);
+  const [hostTagFilters, setHostTagFilters] = useState<HostTagFilter[]>(value ?? []);
```
adamyeats: `value` only seeds the initial state, so any later prop change (query swapped externally, variable duplicated) is ignored. Consider either documenting that this is intentionally uncontrolled or syncing on prop change.

src/datasource/components/VariableQueryEditor.tsx
```diff
@@ -211,3 +274,73 @@ export class ZabbixVariableQueryEditor extends PureComponent<VariableQueryProps,
...
+interface HostTagFilterLoaderProps {
+  datasource: any;
+  group: string;
+  value?: HostTagFilter[];
+  evalTypeValue?: ZabbixTagEvalType;
+  onHostTagFilterChange: (hostTags: HostTagFilter[]) => void;
+  onHostTagEvalTypeChange: (evalType: ZabbixTagEvalType) => void;
+}
+
+const HostTagFilterLoader: React.FC<HostTagFilterLoaderProps> = ({
```
adamyeats, line 295: The tag and value option-building logic here duplicates `loadHostTagOptions` in `MetricsQueryEditor.tsx` almost line for line. Consider extracting a shared helper next to `processHostTags` in `QueryEditor/utils`.

src/datasource/components/VariableQueryEditor.tsx
```diff
@@ -211,3 +274,73 @@ export class ZabbixVariableQueryEditor extends PureComponent<VariableQueryProps,
     );
   }
 }
+
+interface HostTagFilterLoaderProps {
+  datasource: any;
```
adamyeats: `VariableQueryProps` already carries a datasource, and `any` hides mistakes like the speculative `datasource?.templateSrv?.replace` probing below. Consider typing it as the Zabbix datasource.

src/datasource/components/VariableQueryEditor.tsx
```diff
@@ -96,13 +115,42 @@ export class ZabbixVariableQueryEditor extends PureComponent<VariableQueryProps,
...
-    const { queryType, group, host, application, itemTag, item } = this.state;
-    const queryModel = { queryType, group, host, application, itemTag, item, showDisabledItems };
-    this.props.onChange(queryModel, `Zabbix - ${queryType}`);
+    const queryModel = this.buildQueryModel({ showDisabledItems });
+    this.props.onChange(queryModel, `Zabbix - ${queryModel.queryType}`);
+  };
+
+  handleHostTagsChange = (hostTags: HostTagFilter[]) => {
+    const prev = this.state.hostTags ?? [];
+    if (prev.length === hostTags.length && prev.every((f, i) => f === hostTags[i])) {
```
adamyeats: This loop guard relies on the child reusing filter object references between renders, an implementation detail of `HostTagQueryEditor`. Comparing the tag, value and operator fields would be robust to refactors.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.test.ts
```diff
@@ -199,36 +206,34 @@ describe('Zabbix API connector', () => {
...
       expect(zabbixAPIConnector.request).toHaveBeenCalledWith('host.get', {
         output: ['hostid', 'name', 'host'],
         sortfield: 'name',
         selectTags: 'extend',
-        evaltype: 2,
-        tags: [{ tag: 'role', value: 'api', operator: 0 }],
+        selectInheritedTags: 'extend',
       });
+
+      // Both host-a (direct) and host-b (inherited) match `role contains api`/`web` — wait, only
```
adamyeats: This comment is self-correcting narration rather than documentation. Consider replacing it with one line stating the expected outcome.

adamyeats (changes requested):
@MaxDiOrio Thank you! Tested locally and the feature works (ran it against the zabbix74 devenv, and proved all three headline behaviours in Chrome). Some comments below:

src/datasource/datasource.ts
MaxDiOrio: Wired it through rather than hiding the row. `metricFindQuery` now passes `hostTags`/`evaltype` to the Application, Item tag, Item and Item values branches, and `zabbix.ts` threads them down to `getHosts` via `getApps`/`getAllApps`/`getItemTags`/`getAllItems`/`getAllItemsBefore54` (the item paths carry them on the existing `options` bag). Also moved the empty-Host -> `/.*/` coercion out of the Host branch so it applies to every type, and it now keys off filters with a non-empty tag name so a half-filled row no longer flips the host filter to "all hosts". New spec test `should pass hostTags and evaltype for non-Host query types`.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
MaxDiOrio: Good catch, that was a real regression. Replaced the flat `every`/`some` with `matchHostTagFilters()`, which groups filters by tag name and ORs within a group / ANDs across groups for evaltype 0, and ORs everything for evaltype 2. Covered by three new tests under `describe(evaltype semantics)`.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
MaxDiOrio: Fixed. `getHosts` is now `async` and awaits `initVersion()` before building params (`initVersion` is the caching one on the connector; `getVersion` is the raw `apiinfo.version` call). New test `resolves the version before building params so selectInheritedTags is not dropped` starts with `version = undefined` and asserts the param is still set.

src/datasource/datasource.ts
MaxDiOrio: Fixed — host tags are interpolated with a plain `this.templateSrv.replace` for both `tag` and `value` now, with a comment explaining why `replaceTemplateVars` is wrong here. New spec test asserts `$env` -> `production`, not `/^production$/`.

src/datasource/components/QueryEditor/MetricsQueryEditor.tsx
MaxDiOrio: Fixed, and the underlying cause too. `MetricsQueryEditor` now passes `value={query.hostTags}`, and `HostTagQueryEditor` is a fully controlled component — it renders from props with no mirrored state, so there is no mount effect writing an empty list back. New `HostTagQueryEditor.test.tsx` includes `does not write filters back on mount`.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
MaxDiOrio: Fixed — the default arm returns `false` now, with a comment. Test: `matches nothing for an unrecognised operator`.

src/datasource/components/QueryEditor/HostTagQueryEditor.tsx
MaxDiOrio: Went with controlled rather than documenting it as uncontrolled: `value` is now the single source of truth (`const hostTagFilters = value ?? []`), all mutations emit through `onHostTagFilterChange`, and the redundant `hostTagValueDrafts` state is gone (it was always written in lockstep with `filter.value`). That drops both the `useState` and the sync effect — an effect calling `setState` also trips the repo `no-cascading-renders` lint rule. New test `picks up filters changed outside the editor` rerenders with a different `value` and asserts the fields follow.

src/datasource/components/VariableQueryEditor.tsx
MaxDiOrio: Extracted to `buildHostTagOptions(hosts)` in `QueryEditor/utils.ts`, next to `processHostTags`. Both editors call it now; `MetricsQueryEditor.loadHostTagOptions` is down to two lines.

src/datasource/components/VariableQueryEditor.tsx
MaxDiOrio: Typed as `ZabbixDatasource`. That immediately exposed the probe you flagged — `templateSrv` is `private readonly` on the datasource, so `datasource.templateSrv` was never going to resolve. Replaced with `replaceTemplateVars(getTemplateSrv(), group, {})`, which matches how `metricFindQuery` interpolates the group filter.

src/datasource/components/VariableQueryEditor.tsx
MaxDiOrio: Fixed — the guard uses a new `hostTagFiltersEqual()` helper in `QueryEditor/utils.ts` that compares `tag`, `value` and `operator` per element instead of relying on reference identity.

src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.test.ts
MaxDiOrio: Replaced with a single line stating the expected outcome.

MaxDiOrio:
Thanks for the thorough review @adamyeats — all 11 comments addressed in 807b9673, replies inline on each thread.

The three that were actual bugs rather than cleanups:

- **And/Or was a real regression.** The flat `every` broke two filters on the same tag name for the *existing* panel feature. Now grouped: OR within a tag name, AND across names for evaltype 0.
- **The version race** would silently drop `selectInheritedTags` and then CachingProxy would serve that wrong result for 10 minutes. `getHosts` awaits `initVersion()` before building params.
- **`replaceTemplateVars` regex-wrapping** meant a `$var` tag value could never match. Plain `templateSrv.replace` now.

On the first comment I went with wiring the filter through rather than hiding the row — `hostTags`/`evaltype` now reach the Application, Item tag, Item and Item values branches via `zabbix.ts`, so what the editor shows is what actually gets applied.

`HostTagQueryEditor` is now fully controlled (no mirrored state, no mount effect writing an empty list back), which fixes the panel-editor repopulation you spotted and makes external changes work. Note an effect that calls `setState` also trips the repo `no-cascading-renders` lint rule, so controlled was the only clean option here.

Also merged `main` in (b8845f80) — the branch had gone stale: it now uses a changeset instead of a CHANGELOG `Unreleased` section, and `VariableQueryEditor` picks up the local `useAsyncFn` hook since `react-use` was dropped in 6.4.1.

`tsc` clean, `eslint` 0 errors (65 pre-existing deprecation warnings), 244/244 tests pass — up from 177, with new coverage for And/Or grouping, the version race, unknown operators, non-Host propagation, literal interpolation, and a new `HostTagQueryEditor.test.tsx`.

MaxDiOrio:
@adamyeats gentle ping — this has been ready since late July, I think it just slipped off the radar rather than anything being outstanding on my end.

Status:

- **All 11 review comments addressed**, each with a reply on its thread. I've resolved the 8 that GitHub now marks outdated; left the 3 still anchored to live lines open so you can check them yourself.
- **Merged `main` in** (`74772ad0`) — the branch had gone a month stale. Clean merge, no conflicts. This picks up the yarn→npm switch (#2531), the 6.6.0 release, and #2508. Worth noting #2508 touched `VariableQueryEditor.tsx`, the same file this PR changes: its new `VariableQueryEditor.test.tsx` passes unmodified against my version, so the merge is clean behaviourally and not just textually.
- **Signatures sorted.** The earlier "3 of 3 commits could not be verified" report was stale — it ran ~2 minutes before my SSH key was registered. The bot now reports all 4 commits verified.
- **Green locally on the post-merge tree** (node 24.18.0 / npm 11.16.0): `tsc --noEmit` clean, `eslint` 0 errors, **309/309 tests pass**.

One thing I can't do anything about: **no CI has ever run on this PR.** `integration.yml`, `is-compatible.yml` and `push.yaml` all trigger on `pull_request`, but there are zero runs for this branch — I'm an outside contributor, so the runs need a maintainer to approve them. If you've been waiting on a green check before re-reviewing, that's why it never arrived. Would you mind approving the workflow run?

Happy to rebase, split this up, or take another pass at anything if it'd help.

christos-diamantis:
+1

PR #2433, opened 2026-05-29
security hardening: dependency improvements

jcolladokuri:
Adds security hardening to zabbix datasource by:

- removing any unused dependencies
- requiring `>=24.5.0` through `nvmrc` and `package.json`
- hardens the `yarnrc` file by setting required flags and removing `enabledGitRepositories`


package.json
jcolladokuri, line 62: I am removing this in a separate PR since it required migrating some components.

jcolladokuri: https://github.com/grafana/grafana-zabbix/pull/2434

itsmylife (approved):

PR #2434, opened 2026-05-29
security hardening: remove use of @types/grafana package

jcolladokuri:
Throughout zabbix we are using some functions that did not exist as part of an actual published npm package.

We were using `"@types/grafana": "github:CorpGlory/types-grafana"` instead in order to be able to directly import functions from `grafana/app/core/*`. As part of security guidelines, we need to specify a specific pinned dependency version, which for `@types/grafana` this is not possible.

This PR removes the need for `@types/grafana` by migrating functions that were imported from grafana core functionality and instead we are now relying solely on `@grafana/*` packages. All migrated functions were available and supported.


src/test-setup/jest-setup.js
itsmylife, line 63:
```suggestion
```

itsmylife: Duplicate mock

itsmylife (commented):
Can you please check if we need dataMath.js under `test-setup/modules`? If we don't let's remove it too.

src/test-setup/jest-setup.js
jcolladokuri: I could not find any duplicate mocks and when I tried applying your suggestion, github gave me an error because of an updated diff

jcolladokuri:
@itsmylife I checked and datemath.js is indeed no longer needed, thanks for catching this, I have removed it!

zoltanbedi (approved):

PR #2439, opened 2026-06-08
Fix operational data and user macro expansion in problems panel.

### LINKED ISSUE #1832: Operational data display issue in Zabbix problems panel
**Describe the bug**
When using the {ITEM.VALUE} macro in operational data and multiple event generation, Grafana displays the latest item value in the operational data inside the event detail, rather than the value that originally created the problem event.
This behavior is different from Zabbix, where the operational data correctly reflects the value that triggered the problem event.

**Steps to Reproduce**

* Create an item with the following trigger configuration, and populate the item to generate problems.
trigger

latest_data

* Observe the behavior in both Zabbix and Grafana.
zabbix_problems

grafana_problems

In Zabbix, the operational data column works as expected, showing the value that triggered the problem event. Grafana, however, displays the latest item value in the operational data, which may not accurately represent the problem.

Upon inspecting the API calls in the browser's developer mode, I noticed that the calls being made are 'trigger.get' and 'problem.get'. The first one obtains the operational data correctly, but the second one obtains only the last value of the item. 
I suppose that's the problem, use the last value of the item obtained with the problem.get method instead the operational data obtained with the trigger.get call.

grafana_problem get

grafana_trigger get

Note: If you enable the operational data column in the Zabbix problems panel, it will correctly display the operational data.

grafana_opdata_column

**Software versions**

| 10.0.1   | 5.0.30 | 4.3.1     |
| 10.1.4   | 5.0.30 | 4.4.1     |
| 10.1.4   | 5.0.42 | 4.4.9     |
| 11.0   | 5.0.42 | 4.4.9     |
| 11.0   | 6.0.30 | 4.4.9     |

zoltanbedi:
Expand {ITEM.VALUE} in opdata using historical item values at problem creation time, and resolve user-defined {$MACRO} placeholders in problem description, comments, and operational data.

Fixes #1832

### DIFF
--- a/.changeset/fix-opdata-user-macros.md
+++ b/.changeset/fix-opdata-user-macros.md
@@ -1,0 +1,5 @@
+---
+'grafana-zabbix': patch
+---
+
+Fix operational data macro expansion and user macro support in problems panel
--- a/src/datasource/problemsHandler.test.ts
+++ b/src/datasource/problemsHandler.test.ts
@@ -95,12 +95,7 @@ describe('expandUserMacros', () => {
   const globalMacro = (macro: string, value: string) => ({ macro, value });
 
   it('replaces a host macro with its value', () => {
-    const result = expandUserMacros(
-      'Contact: {$CONTACT}',
-      [hostMacro('{$CONTACT}', 'Pepe', '10')],
-      [],
-      ['10']
-    );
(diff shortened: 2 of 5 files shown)


kentped:
@itsgareth Hi Team

I am checking if there is any updates here. This is the continuation of a fix to a support ticket opened in April

Thanks!

alyssajoyner (approved):
LGTM

Issue #2427, opened 2026-05-25
[CRITICAL BUG] 6.3.1 / 6.3.2: N+1 history.get queries per problem cause severe DB overload and PHP-FPM timeouts in large environments


christos-diamantis:
<html><head></head><body>
<h2>Summary</h2>
<p>Versions <strong>6.3.1</strong> and <strong>6.3.2</strong> introduced a per-problem <code>history.get</code> API call pattern that generates a massive number of individual <code>SELECT</code> queries against the Zabbix <code>history_uint</code> (and <code>history</code>) tables. In large environments with many hosts and items, this causes:</p>
<ul>
<li>PHP-FPM workers timing out (10–13+ seconds per <code>api_jsonrpc.php</code> request)</li>
<li>PostgreSQL saturated with slow <code>history_uint</code> range scans averaging <strong>2,000–6,000ms each</strong></li>
<li>Zabbix web UI becoming unresponsive or unusable</li>
<li>Cascading load on the database cluster</li>
</ul>
<p><strong>Rolling back to 6.3.0 immediately resolves all symptoms.</strong> This is a confirmed regression introduced in 6.3.1.</p>
<hr>
<h2>Environment</h2>

Component | Details
-- | --
Plugin version (broken) | alexanderzobnin-zabbix-app 6.3.1 and 6.3.2
Plugin version (working) | alexanderzobnin-zabbix-app 6.3.0 ✅
Zabbix version | 7.x
Database | PostgreSQL 18 with TimescaleDB, Patroni HA cluster (3 nodes)
Zabbix frontend | 2-node HA, PHP-FPM 8.x, pm = static, 150 workers per node
Grafana nodes | 4 nodes behind F5 LB
Scale | Very large environment: thousands of hosts and millions of items

<hr>
<h2>Suggested Fix</h2>
<h3>Option 1 — Batch the <code>history.get</code> calls (minimum viable fix)</h3>
<p>Instead of issuing one API call per problem, the plugin should:</p>
<ol>
<li>Collect all <code>(itemid, timestamp)</code> pairs for all visible problems</li>
<li>Issue a <strong>single batched <code>history.get</code> request</strong> covering all itemids and the full timestamp range needed</li>
<li>Map results back to individual problems client-side</li>
</ol>
<p>This is the same approach that worked correctly in 6.3.0 for operational data — the fix in 6.3.1 should have preserved batching semantics while correcting the value-at-timestamp logic.</p>
[...]

PR #2458, opened 2026-06-28
ci: add Playwright e2e environment and tests

adamyeats:
Adds a canonical e2e testing environment and the Playwright suites that run against it. **Stacked on #2460** (the #2427 fix) — its base is the fix branch because the `history.get` e2e test needs the fix to pass. **Merge #2460 first**; GitHub then retargets this PR to `main`.

## Environment

- **`docker-compose.yml`** — canonical dev/e2e stack: Postgres + Zabbix server/web + a one-shot **`e2e-data-loader`** + Grafana (extends the base compose, `depends_on` the loader via `service_completed_successfully`, so tests never run against an empty backend). Zabbix version is switchable via **`ZABBIX_VERSION`**, defaulting to **7.0 (latest LTS)**.
- **`tests/fixtures/*.sql`** — loaded in lexicographic order via `psql`, so feature fixtures can sit alongside `seed.sql` without merge conflicts. `seed.sql` creates a host + items + triggers and two open problems (one recent, one backdated ~30 days) + history.
- **`provisioning/`** — moved to the repo root (the default `provisioningRootDir`, mounted by the base compose); data source at `uid: zabbix-e2e`. Un-ignored in `.gitignore`.
- **`push.yaml`** — `run-playwright: true` runs `tests/e2e` against `docker-compose.yml` via the shared workflow.
- **`.trufflehog.yml`**, `playwright.config.ts`, and docs (`CONTRIBUTING.md`, `tests/e2e/README.md`, `devenv/README.md`).
- **devenv stacks** redirect their Docker-generated provisioning mount points to a gitignored `tmp/` dir, so running a per-version `devenv/zabbixNN` stack no longer pollutes the now-tracked `provisioning/`.

## Tests (`tests/e2e`)

- **`configEditor.spec.ts`** — rendering (smoke + sections), the provisioned data source loads its connection settings, and Save & test (passes against a reachable Zabbix; errors when unreachable). The health-check tests create an **ephemeral** data source — provisioned URLs are read-only — and tear it down.
- **`queryEditor.spec.ts`** — rendering (query editor + metrics filter fields) and a fixture-data Problems query that returns the seeded problems from the real backend.
- **`problemsHistoryBounded.spec.ts`** — the #2427 test: with the option disabled (default) the Problems panel sends no `history.get`; with it enabled the `history.get` is bounded (`limit: 50000`, window span ≤ ~5h) against the backdated-problem fixture.

## Local DX

`yarn e2e` runs against a single shared backend, so the worker count is capped (`workers: 2` locally, `1` in CI). Playwright's default (≈ half the CPU cores) overloads the backend and causes timeouts — the cap makes `yarn e2e` reliable with no `--workers` flag.

## Versions / transport

Per-version `devenv/` stacks (TLS + basic auth on 5.0/6.0/7.0, proxy on 5.0) and the `compatibility-*`/integration Go workflows are left untouched — they own version + transport/auth coverage. The frontend e2e is transport-agnostic, so it runs against a single apache-backed version (switchable via `ZABBIX_VERSION`).

## Validation

Full suite passes serially and at the local worker cap; CI Playwright matrix is green across Grafana 11.6 → 13.1, react19, and nightly. typecheck / lint / cspell clean.


zoltanbedi (dismissed):
LGTM

zoltanbedi (approved):

PR #2461, opened 2026-06-28
ci: split integration tests (matrix) from the Grafana API compatibility check

adamyeats:
Disambiguates two things that were conflated by the old `compatibility-NN.yml` naming.

## 1. Integration tests → single matrix (`integration.yml`)

The five `compatibility-NN.yml` workflows were **misnamed**: they run the Go API integration tests (`pkg/zabbixapi`) against a real Zabbix instance, not a Grafana-API compatibility check. Consolidated into one `integration.yml` that runs them concurrently via a Zabbix-version matrix:

| Zabbix | devenv | env gate | URL |
|---|---|---|---|
| 5.0, 6.0 | `devenv/zabbix50`,`60` | `INTEGRATION_TEST50/60` | `https://localhost/...` (nginx TLS + basic auth) |
| 7.0, 7.2, 7.4 | `devenv/zabbix70`,`72`,`74` | `INTEGRATION_TEST70/72/74` | `http://localhost:8188/...` (apache) |

`fail-fast: false`; each version's Go test self-gates on its own `INTEGRATION_TESTNN` var (passed via `env`, not interpolated into `run`, to stay zizmor-clean).

## 2. Add the real compatibility check (`is-compatible.yml`)

Now that "compatibility" is no longer overloaded, add [`grafana/plugin-actions/is-compatible`](https://github.com/grafana/plugin-actions/tree/main/is-compatible) (levitate) — which detects use of deprecated/removed **Grafana frontend APIs** and comments on the PR. This is the established meaning of "compatibility" for plugins; it's distinct from the Zabbix integration tests.

Builds the plugin then runs levitate against `@grafana/data,@grafana/ui,@grafana/runtime,@grafana/schema`; comments only on incompatibility (`skip-comment-if-compatible: yes`), non-blocking (`fail-if-incompatible: no`).

Both workflows validated with `actionlint`.

## ⚠️ Required status checks

Check names change: `compatibility-check` (×5) → **`integration (Zabbix <version>)`** (×5), plus a new **`compatibilitycheck`**. Branch-protection required-status-check settings will need updating.


zoltanbedi (approved):
Nice!

PR #2463, opened 2026-06-29
Support item tag values as alias tokens

### LINKED ISSUE #2449: [Zabbix] Support item_tag values as tokens in Replace Alias (and other alias/display functions)
In the Zabbix data source, `item_tag` values can be used during query selection (filtering which items to fetch), but they cannot be referenced from display/alias functions such as Replace Alias, Set Alias, or Alias By Regex. Only $hostname, $item, $key, and the existing built-in tokens are resolvable in the alias output.

This forces users who organize their items with tags (a Zabbix-recommended practice) to fall back to item_key-based aliases, which produce legend strings like vfs.fs.size[/,used] instead of a human-readable label.

Allow item_tag values to be referenced as tokens in alias-formatting functions, with parity to how $hostname and $key work today.

Proposed syntax (open to alternatives):
- `$tag.<tag_name> `→ resolves to the value of the named tag on the item (e.g. $tag.service → checkout).
- `$tag.<tag_name>` returns an empty string if the tag is not present on the item.
- When multiple values exist for the same tag name on a single item, the values are joined with , (or the plugin's existing multi-value convention — whichever the maintainers prefer).

Functions that should support the new token at minimum:

- `setAlias`
- `replaceAlias`
- `aliasByRegex / aliasByMetric`
- any other function whose output is the rendered legend string

- Hard-coded Replace Alias per item key — current workaround. Brittle, doesn't scale, breaks on new items.
- Renaming the underlying Zabbix items — disruptive, touches operational naming conventions, often not feasible.
- Post-processing the exported PDF — fragile and out of scope of the report workflow.

iwysiu:
## Summary
- Add `$__zbx_item_tag_<name>` alias tokens for Zabbix item tag values in `setAlias` and `replaceAlias`
- Populate tag scopedVars in both Go and TS response handlers, with sanitized tag names and sorted multi-value joins
- Fix `replaceAlias` to read `config.custom.scopedVars` and copy scopedVars into frontend-built DataFrames
- Fetch item tags in `GetItemsByIDs` on Zabbix 5.4+

Closes #2449

## Test plan
- [ ] `go test ./pkg/datasource/ -run 'TestSanitizeTagName|TestAddItemTagScopedVars|TestSeriesToDataFrameItemTagScopedVars'`
- [ ] `yarn jest --testPathPatterns='utils.spec|dataProcessor.spec|responseHandler.spec'`
- [ ] In Grafana, query items with tags and use `setAlias($__zbx_item_tag_<name>)` to verify legend text
- [ ] Confirm tags with no value resolve to empty string and absent tags leave the literal token
- [ ] Verify item-ID queries include tags on Zabbix 5.4+

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/sources/functions.md
+++ b/docs/sources/functions.md
@@ -331,7 +331,7 @@ Alias functions change the display names of time series. The following template 
 | `$__zbx_host_name` | Visible name of the host. |
 | `$__zbx_host` | Technical name of the host. |
 | `$__zbx_host_id` | ID of the host. |
-| `$__zbx_item_tag_<name>` | Value of the item tag with the given name. Tag names are sanitized to alphanumeric characters and underscores (for example, `App Name` becomes `$__zbx_item_tag_App_Name`). A tag that exists but has no value resolves to an empty string; a tag that is not present on the item is left unresolved (the literal token remains in the output). When multiple values exist for the same tag name, they are joined with `, `. Requires Zabbix 5.4 or later. |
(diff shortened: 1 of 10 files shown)


k-munoz:
I was able to test it!

Screenshot 2026-07-01 at 10 29 21 AM

lwandz13 (approved):
Looks great - thank you!!

mattvella07 (approved):
LGTM

PR #2467, opened 2026-06-30
feat(problems): filter problem names at the Zabbix API source

### LINKED ISSUE #2466: Problems panel: problem-name filtering is client-side, causing huge payloads and timeouts in large environments
### What happened

When a Problems query uses the **Problem** (name) filter, the filter is
applied **client-side, after** the plugin has already fetched every problem
in the group/host scope. The data source:

1. calls `problem.get` with `output: extend` + `selectAcknowledges` +
   `selectTags` + `selectSuppressionData` for **all** problems in scope,
2. calls `trigger.get` for **every** returned problem, and only then
3. filters by problem name in the browser.

### Impact

In a large enterprise setup (multiple environments, many active alerts) a
dashboard with several Problems panels - each pulling everything and then
filtering - produces very large responses. This:

- hits gRPC message-size limits between the plugin backend and Grafana, and
- causes `context deadline exceeded` / query timeouts.

There's also a secondary issue: the query `limit` is applied by the API
**before** the name filter, so increasing the limit to find named problems
just enlarges the payload and can still miss matches.

### Expected

Problem-name filtering should happen **at the source** (Zabbix API) so only
matching problems are transferred, with regex still supported.

### Notes

Zabbix `problem.get` / `event.get` support a `search` parameter (case-
insensitive SQL `LIKE`, optional `*` wildcards, `searchByAny` for OR). It
cannot evaluate regular expressions, so regex needs a literal-substring
pre-filter at the source plus a precise client-side pass.

Environment: Zabbix 7.x, large multi-environment installation.

christos-diamantis:
## Summary

Pushes the Problems-query **name filter** down to the Zabbix API
(`problem.get` / `event.get` `search`) instead of fetching every problem in
scope and filtering in the browser. This drastically cuts the data
transferred in large environments (avoiding gRPC limits / timeouts) and
makes `limit` apply to name-matched problems.

Fixes #2466 

## How the filter maps to the API

| Filter type | Pushed to Zabbix | Client-side pass |
|---|---|---|
| Plain text `High CPU load` | `search:{name}` (substring) | exact match |
| Wildcard `CPU*` | `search:{name}` + `searchWildcardsEnabled` | wildcard match (**new**) |
| Regex `/CPU load.*/` | `search:{name:"CPU load"}` (guaranteed literal) | full regex |
| Alternation `/(A\|B\|C)/` | `search:{name:[A,B,C]}` + `searchByAny` | full regex |
| Look-around `/^(?!...)/` | *(not pushed)* | full regex |

The server step is always a **superset** (case-insensitive substring); the
client-side regex/exact pass tightens it, so results are identical to before
- just cheaper. When a regex can't be safely narrowed (a non-literal
alternation branch, or any look-around construct), it falls back to
client-side filtering only, so correctness is never traded for speed. In
particular, a literal inside a **negative** look-ahead is deliberately not
pushed (doing so would return exactly the rows the user wants excluded).

## Testing

- New unit tests cover plain/wildcard/regex/alternation/look-around mapping,
  the connector params, and client-side matching (including the alternation
  and negative-look-ahead examples).

### DIFF
--- a/src/datasource/datasource.ts
+++ b/src/datasource/datasource.ts
@@ -591,6 +591,10 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery, 
       recent: showProblems === ShowProblemTypes.Recent,
       minSeverity: target.options?.minSeverity,
       limit: target.options?.limit,
+      // Filter problem names at the source (Zabbix API) instead of fetching all
+      // problems and filtering client-side. The precise client-side filter still
+      // runs afterwards (see problemsHandler.filterTriggersPre).
+      problemName: target.trigger?.filter,
     };
 
     if (tags && tags.length) {
@@ -818,6 +822,8 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery, 
       valueFromEvent: true,
       timeFrom,
       timeTo,
+      // Narrow problem names at the source; the precise filter still runs below.
+      problemName: annotation.trigger?.filter,
     };
 
     if (annotation.options.minSeverity) {
@@ -835,14 +841,8 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery, 
       .then((problems) => {
         // Filter triggers by description
         const problemName = annotation.trigger.filter;
-        if (utils.isRegex(problemName)) {
-          problems = _.filter(problems, (p) => {
-            return utils.buildRegex(problemName).test(p.description);
-          });
-        } else if (problemName) {
-          problems = _.filter(problems, (p) => {
-            return p.description === problemName;
-          });
+        if (problemName) {
+          problems = _.filter(problems, (p) => utils.matchesProblemName(p.description ?? '', problemName));
         }
 
         // Hide acknowledged events if option enabled
--- a/src/datasource/datasource.ts
+++ b/src/datasource/datasource.ts
@@ -839,10 +839,11 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery, 
     return this.zabbix
       .getProblemsHistory(groupFilter, hostFilter, appFilter, proxyFilter, problemsOptions)
       .then((problems) => {
-        // Filter triggers by description
+        // Filter by problem name — matched on `name`, the same field the server-side
+        // `search` narrowing filters on (see problemsHandler.filterTriggersPre).
         const problemName = annotation.trigger.filter;
         if (problemName) {
-          problems = _.filter(problems, (p) => utils.matchesProblemName(p.description ?? '', problemName));
+          problems = _.filter(problems, (p) => utils.matchesProblemName(p.name ?? '', problemName));
         }
 
         // Hide acknowledged events if option enabled
--- a/src/datasource/problemsHandler.ts
+++ b/src/datasource/problemsHandler.ts
@@ -186,15 +186,7 @@ export function filterTriggersPre(triggerList, replacedTarget) {
 }
 
 function filterTriggers(triggers, triggerFilter) {
-  if (utils.isRegex(triggerFilter)) {
-    return _.filter(triggers, (trigger) => {
-      return utils.buildRegex(triggerFilter).test(trigger.description);
-    });
-  } else {
-    return _.filter(triggers, (trigger) => {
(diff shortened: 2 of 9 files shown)


src/datasource/problemsHandler.ts
```diff
@@ -186,15 +186,7 @@ export function filterTriggersPre(triggerList, replacedTarget) {
...
-  if (utils.isRegex(triggerFilter)) {
-    return _.filter(triggers, (trigger) => {
-      return utils.buildRegex(triggerFilter).test(trigger.description);
-    });
-  } else {
-    return _.filter(triggers, (trigger) => {
-      return trigger.description === triggerFilter;
-    });
-  }
+  return _.filter(triggers, (trigger) => utils.matchesProblemName(trigger.description ?? '', triggerFilter));
```
mattvella07: Consider matching on `trigger.name` instead of `trigger.description` here. On the event.get path, description is the trigger name (macros unexpanded) but the search you push narrows on the event name (macros expanded). Since the server filters first, rows can be dropped before this check runs — so triggers with macros in their name (e.g. High CPU on {HOST.NAME} → High CPU on server1) can lose results that used to show.

`trigger.name` is populated on both paths, and matches description on the problem.get path, so it should be safe.

src/datasource/datasource.ts
```diff
@@ -835,14 +841,8 @@ export class ZabbixDatasource extends DataSourceWithBackend<ZabbixMetricsQuery,
...
-        if (utils.isRegex(problemName)) {
-          problems = _.filter(problems, (p) => {
-            return utils.buildRegex(problemName).test(p.description);
-          });
-        } else if (problemName) {
-          problems = _.filter(problems, (p) => {
-            return p.description === problemName;
-          });
+        if (problemName) {
+          problems = _.filter(problems, (p) => utils.matchesProblemName(p.description ?? '', problemName));
```
mattvella07: Same as `problemsHandler.ts`, consider matching on `p.name`

src/datasource/utils.ts
```diff
@@ -231,6 +231,258 @@ export function buildRegex(str) {
...
+      // Character class: its contents are alternatives, not a guaranteed literal — skip it.
+      flush(false);
+      i++;
+      while (i < pattern.length && pattern[i] !== ']') {
+        if (pattern[i] === '\\') {
+          i++;
+        }
+        i++;
+      }
+    } else if (ch === '?' || ch === '*' || ch === '{') {
```
mattvella07, line 331: `flush(true)` only drops the single preceding character, so an optional group slips through as a "guaranteed" literal.

For `/(Critical failure)?CPU load/` this extracts "Critical failure" and pushes it to search, which drops any plain "…CPU load" problem the regex actually matches. 

It might be worth retracting the whole preceding run when ?/*/{ follows a ) (or ]), rather than just the last char — and adding a test for /(Critical failure)?CPU load/ to cover it.

src/datasource/problemsHandler.ts
christos-diamantis: Pretty nice catch! Fixed on the latest commit.

src/datasource/utils.ts
christos-diamantis: done on the latest commit

src/datasource/datasource.ts
christos-diamantis: done on the latest commit

christos-diamantis:
@mattvella07 than you very much for the review! Fixed all the issues and added new tests. Specifically:

- filter problem by name instead of description. same way the search does it server side
- added tests for filtering triggers using p.name
- filter event by name instead description cause name has macros expanded on the creation
- guaranteedLiteralForBranch now handles (/) explicitly with a mark stack: each ( records runs.length when ) is followed by ?, *, or {, every run committed inside that group is retracted. + after ) keeps the group's runs (guaranteed >= 1 occurrence), and unquantified groups keep theirs
- 7 new quantified groups cases: ?/*/{0,2} retraction, multi-run retraction, nested groups, + kept, unquantified kept, plus 2 buildProblemNameSearchParams cases (optional group -> remaining literal, only-optional-group -> no narrowing)
- /(Critical failure)?CPU load/ test also done on the API Connector test to verify the correct parameter is send as search to the Zabbix API

mattvella07 (approved):
LGTM

Issue #1295, opened 2021-09-09
Zabbix API request ERROR "EOF"


CLU91:
**Describe the bug**

Throughout the day it happens multiple times that the Grafana Dashboard will crash for a few seconds with the message:
Post \"https://zabbix.domain-name.de/api_jsonrpc.php\": EOF
After that we need to refresh the webpage. Sometimes it's not happening at all, sometimes it's happening quite a lot throughout the day. 

**Expected behavior**

The Zabbix Plugin not to crash sporadically

**Screenshots**

No Screenshot, but the ERROR MSG from the Grafana.log:
t=2021-09-09T01:31:34+0200 lvl=eror msg="Zabbix API request error" logger=plugins.backend pluginId=alexanderzobnin-zabbix-datasource error="Post \"https://zabbix.domain-name.de/api_jsonrpc.php\": EOF"
t=2021-09-09T01:31:34+0200 lvl=eror msg="Request Completed" logger=context userId=2 orgId=1 uname=monitoring method=POST path=/api/datasources/1/resources/zabbix-api status=500 remote_addr=192.168.42.105 time_ms=3 size=119 referer="https://grafana.domain-name.de/d/WwCQTgm7z/druckerserver?kiosk&orgId=1&refresh=5s"

**Network data**

/

**Software versions**

|               Grafana       |    Zabbix    | Grafana-Zabbix Plugin |
| ---------------------- | ------------ | ------------------------ |
| v8.0.4 (81fef5b4b4)   |    5.0.15     |                4.1.5              |

alexanderzobnin:
Hm, probably, it's golang-related issue, I have to set `req.Close = true` in order to avoid it.

CLU91:
UPDATE: I updated the Plugin Version to 4.2.2 and now i have one more issue in addition. I want to monitor the Zabbix Agent Availability. The value from Zabbix is 1 (UP). I also tried Zabbix Agent Ping which is also the value 1 for the status "up".

Both sporadically jumping to the value 0.500. Since i set value 1 to green and 0 to red, i have several servers jumping from green to red all the time and back.

alexanderzobnin:
@CLU91 could you open this as a separate issue, this is another problem (or just panel misconfiguration)?

alexanderzobnin:
I added fix for this, let me know if it works or not (after 4.2.3 release).

CLU91:
Hi Alexander,

where to get the release 4.2.3? The error is appearing every minute now...

PR #2496, opened 2026-07-20
fix: connection storm against Zabbix web frontend causing 502/503 und…

### LINKED ISSUE #2495: fix connection storm 503 under load
**Describe the bug**

The datasource backend generates far more HTTP requests to the Zabbix web
frontend than it needs to, which under normal dashboard load (several
panels, autorefresh, template-variable dropdowns) can exhaust php-fpm/nginx
connection limits on modest self-hosted Zabbix installs and surface as
widespread 502/503 errors in Grafana.

Two issues compound each other in `pkg/zabbix/zabbix.go` and
`pkg/zabbixapi/zabbix_api.go`:

1. `Zabbix.Request()` calls `GetVersion()` → `GetFullVersion()`
   unconditionally on **every** API call (metrics queries, the metric-picker
   autocomplete for groups/hosts/items/apps, health checks - everything).
   `GetFullVersion()` always issues a live `apiinfo.version` request,
   bypassing the `ZabbixCache` that already exists and is used for
   `host.get`/`item.get`/etc. So every logical operation the plugin performs
   silently becomes 2 HTTP requests instead of 1.

2. `zabbixapi.makeHTTPRequest()` sets `req.Close = true` on every request,
   preventing Go from reusing pooled connections. This was added in 2021
   (commit `a69c866b`, "Do not reuse connection to prevent EOF error", #1295)
   to work around a narrow race on a stale idle connection, but disables
   keep-alive for literally every request since, unconditionally.

Combined, a dashboard with N queries/panels opens roughly **2N fresh TCP
connections** to the Zabbix web frontend per load or refresh, instead of ~N
requests over a couple of pooled connections. On a Zabbix install with a
small php-fpm pool (a common default for self-hosted setups), this reliably
exhausts the pool/connection limits under everyday use, and Zabbix's web
frontend starts answering with 502/503.

**Expected behavior**

The plugin should send one HTTP request per logical Zabbix API call
(reusing the already-detected/cached API version), and should reuse TCP
connections normally like any other well-behaved HTTP client, so a normal
dashboard doesn't generate an outsized number of connections to Zabbix.

**Reproduction / measurement**

Load-tested against a real Zabbix 7.0 (zabbix-web-nginx-pgsql with a small,
6-worker static php-fpm pool) by replaying the plugin's exact request
pattern directly against `api_jsonrpc.php`:

| | HTTP requests | Time (concurrency=40, 10 queries/worker) |
| --- | --- | --- |
| current behavior (2 requests/query, new connection per request) | 800 | 20.16s |
| version cached + connections reused (1 request/query) | 400 | 10.19s |

At higher concurrency (~120+ against this pool size) the current behavior
degrades to ~98% request failures (timeouts) against the same backend that
stays fully healthy under the fixed behavior.

**Software versions**

| 12.3.1  | 7.0.28 | 6.5.0 (introduced 2021, unchanged since) |

I have a fix ready (cache the version via the existing `ZabbixCache`, drop
`req.Close=true` in favor of a single safe retry on the specific
pre-response network error #1295 was about, plus a couple of related
hardening fixes) and will open a PR referencing this issue.

JvstTeo:
Fixes #2495

## Problem

Every backend API call implicitly re-fetched `apiinfo.version` on every
`Zabbix.Request()` instead of reusing the existing `ZabbixCache`, doubling
the number of HTTP requests per logical query. Each of those requests then
forced `req.Close = true`, disabling connection reuse entirely - a 2021
workaround (`a69c866b`, #1295) for a narrow stale-connection race, applied
far too broadly and left unchanged since.

Together: a dashboard with N panels/queries opened ~2N fresh TCP connections
to the Zabbix web frontend per load/refresh instead of ~N requests over a
couple of pooled connections. On modest self-hosted Zabbix installs this
exhausts php-fpm/connection limits under everyday use and surfaces as
widespread 502/503 in Grafana.

## Fix

- `pkg/zabbix/methods.go`: `GetFullVersion` now caches the result via the
  existing `ZabbixCache` (the same mechanism already used for
  `host.get`/`item.get`/etc.) instead of hitting the API on every call.
- `pkg/zabbixapi/zabbix_api.go`: removed `req.Close = true`. In its place,
  `makeHTTPRequest` retries once when the specific pre-response network
  error from #1295 (`io.EOF` / `io.ErrUnexpectedEOF` / `ECONNRESET`) occurs -
  safe even for non-idempotent calls, since by definition the server never
  saw the request in that case.
- `pkg/datasource/resource_handler.go`: `ZabbixAPIHandler` (backs the query
  editor's group/host/item/app autocomplete) is now bounded by the
  configured query timeout, matching the pattern already used in
  `QueryData`, so a slow/overloaded Zabbix can't hang it indefinitely.
  Also fixed `writeError` always responding with 500 regardless of the
  intended status code, and a nil-pointer panic when the request body is
  empty.
- `src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts`: the
  frontend now retries a transient 502/503/504 from the resource endpoint
  (with backoff) before surfacing an error - this is the shared path for
  metric-picker autocomplete and all frontend-mode queries (triggers,
  problems, text metrics, user macros, IT services).

The main metrics query pipeline (`DataSourceWithBackend.query()`) is
intentionally untouched - the root cause is fixed server-side, and
overriding that pipeline for retries felt like unnecessary risk for this PR.

## Testing

- `go test ./pkg/...` - full suite green, plus new tests:
  `TestVersionIsCachedAcrossRequests`, `TestRetryOnTransientConnectionError`,
  `TestRetryOnTransientConnectionErrorGivesUp`,
  `TestNonRetryableErrorIsNotRetried`.
- `yarn jest src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.test.ts`
  - full suite green, plus 3 new tests covering the retry/backoff/give-up
  behavior.
- `yarn typecheck` / `yarn eslint` on the changed files - clean (pre-existing
  `.toPromise()`/`withCredentials` deprecation warnings only, unrelated to
  this change).
- Load-tested against a real Zabbix 7.0 backend (zabbix-web-nginx-pgsql,
  6-worker static php-fpm pool) by replaying the plugin's exact old vs new
  request pattern directly against `api_jsonrpc.php`: ~2x fewer requests and
  ~2x faster at moderate concurrency (40 workers), and the old pattern
  degrades to ~98% failures at higher concurrency (~120+) against a backend
  that stays fully healthy under the fixed pattern. Happy to share the
  load-test script if useful.

### DIFF
--- a/pkg/datasource/resource_handler.go
+++ b/pkg/datasource/resource_handler.go
@@ -1,6 +1,7 @@
 package datasource
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"net/http"
@@ -69,7 +70,18 @@ func (ds *ZabbixDatasource) ZabbixAPIHandler(rw http.ResponseWriter, req *http.R
 
 	apiReq := &zabbix.ZabbixAPIRequest{Method: reqData.Method, Params: reqData.Params}
 
-	result, err := dsInstance.ZabbixAPIQuery(req.Context(), apiReq)
+	// Bound resource calls (metric-picker autocomplete: groups/hosts/items/apps)
+	// with the same query timeout used for regular metric queries in
+	// QueryData (see datasource.go), so a slow/overloaded Zabbix can't hang
+	// this call indefinitely and trip Grafana's own outer gateway timeout.
+	queryTimeout := dsInstance.Settings.QueryTimeout
+	if queryTimeout <= 0 {
+		queryTimeout = 60 * time.Second
+	}
+	resourceCtx, cancel := context.WithTimeout(req.Context(), queryTimeout)
+	defer cancel()
+
+	result, err := dsInstance.ZabbixAPIQuery(resourceCtx, apiReq)
 	if err != nil {
 		ds.logger.Error("Zabbix API request error", "error", err)
 		writeError(rw, http.StatusInternalServerError, err)
@@ -153,8 +165,12 @@ func writeResponse(rw http.ResponseWriter, result *ZabbixAPIResourceResponse) {
 func writeError(rw http.ResponseWriter, statusCode int, err error) {
 	data := make(map[string]interface{})
 
-	data["error"] = "Internal Server Error"
-	data["message"] = err.Error()
+	data["error"] = http.StatusText(statusCode)
+	if err != nil {
+		data["message"] = err.Error()
+	} else {
+		data["message"] = "empty request body"
+	}
 
 	var b []byte
 	if b, err = json.Marshal(data); err != nil {
@@ -163,7 +179,7 @@ func writeError(rw http.ResponseWriter, statusCode int, err error) {
 	}
 
 	rw.Header().Add("Content-Type", "application/json")
-	rw.WriteHeader(http.StatusInternalServerError)
+	rw.WriteHeader(statusCode)
 
 	_, err = rw.Write(b)
 	if err != nil {
--- a/pkg/datasource/resource_handler.go
+++ b/pkg/datasource/resource_handler.go
@@ -3,11 +3,13 @@ package datasource
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"time"
 
 	"github.com/alexanderzobnin/grafana-zabbix/pkg/zabbix"
+	"github.com/alexanderzobnin/grafana-zabbix/pkg/zabbixapi"
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
@@ -84,7 +86,11 @@ func (ds *ZabbixDatasource) ZabbixAPIHandler(rw http.ResponseWriter, req *http.R
 	result, err := dsInstance.ZabbixAPIQuery(resourceCtx, apiReq)
 	if err != nil {
 		ds.logger.Error("Zabbix API request error", "error", err)
-		writeError(rw, http.StatusInternalServerError, err)
+		// Surface the real upstream status (e.g. 502/503/504 from an
+		// overloaded Zabbix) instead of always reporting 500, so the
+		// frontend's retry-on-transient-error logic (which keys off the
+		// actual HTTP status of this response) has something to match.
+		writeError(rw, statusCodeFromError(err, http.StatusInternalServerError), err)
 		return
 	}
 
@@ -160,6 +166,16 @@ func writeResponse(rw http.ResponseWriter, result *ZabbixAPIResourceResponse) {
 	if err != nil {
 		log.DefaultLogger.Warn("Error writing response")
 	}
+}
+
+// statusCodeFromError extracts the upstream HTTP status code carried by a
+// *zabbixapi.StatusError, if any, falling back to fallback otherwise.
+func statusCodeFromError(err error, fallback int) int {
+	var statusErr *zabbixapi.StatusError
+	if errors.As(err, &statusErr) {
+		return statusErr.StatusCode
+	}
+	return fallback
 }
 
 func writeError(rw http.ResponseWriter, statusCode int, err error) {
--- a/pkg/zabbix/methods.go
+++ b/pkg/zabbix/methods.go
@@ -537,7 +537,25 @@ func (ds *Zabbix) GetValueMappings(ctx context.Context) ([]ValueMap, error) {
 	return valuemaps, err
 }
 
+// GetFullVersion returns the Zabbix API version, e.g. "6.4.5".
+//
+// The version rarely changes for a running Zabbix instance, but Request()
+// (see zabbix.go) checks it on every single API call to pick the right
+// request dialect. Without caching, that turns every logical Zabbix API
+// call the plugin makes into two HTTP requests instead of one, which under
+// real dashboard load (many panels, autorefresh, several users) is enough
+// to exhaust the connection/worker limits of the Zabbix web frontend and
+// makes it start answering with 502/503. Cache the result using the same
+// ZabbixCache already used for host.get/item.get/etc. so the version is
+// only re-fetched once per cache TTL.
 func (ds *Zabbix) GetFullVersion(ctx context.Context) (string, error) {
+	versionReq := &ZabbixAPIRequest{Method: "apiinfo.version"}
+	if cached, ok := ds.cache.GetAPIRequest(versionReq); ok {
+		if version, ok := cached.(string); ok && version != "" {
+			return version, nil
+		}
+	}
+
 	result, err := ds.request(ctx, "apiinfo.version", ZabbixAPIParams{})
 	if err != nil {
 		return "", err
@@ -548,7 +566,9 @@ func (ds *Zabbix) GetFullVersion(ctx context.Context) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	
+
+	ds.cache.SetAPIRequest(versionReq, version)
+
 	return version, nil
 }
 
--- a/pkg/zabbix/methods.go
+++ b/pkg/zabbix/methods.go
@@ -545,12 +545,15 @@ func (ds *Zabbix) GetValueMappings(ctx context.Context) ([]ValueMap, error) {
 // call the plugin makes into two HTTP requests instead of one, which under
 // real dashboard load (many panels, autorefresh, several users) is enough
 // to exhaust the connection/worker limits of the Zabbix web frontend and
-// makes it start answering with 502/503. Cache the result using the same
-// ZabbixCache already used for host.get/item.get/etc. so the version is
-// only re-fetched once per cache TTL.
+// makes it start answering with 502/503. Cache the result in a dedicated
+// cache (not the ZabbixAPIRequest-keyed ZabbixCache used for
+// host.get/item.get/etc.) so the version is only re-fetched once per cache
+// TTL, and so an unrelated apiinfo.version call routed through Request()
+// can never read this string value back expecting a *simplejson.Json.
+const versionCacheKey = "version"
+
 func (ds *Zabbix) GetFullVersion(ctx context.Context) (string, error) {
-	versionReq := &ZabbixAPIRequest{Method: "apiinfo.version"}
-	if cached, ok := ds.cache.GetAPIRequest(versionReq); ok {
+	if cached, ok := ds.versionCache.Get(versionCacheKey); ok {
 		if version, ok := cached.(string); ok && version != "" {
 			return version, nil
 		}
@@ -567,7 +570,7 @@ func (ds *Zabbix) GetFullVersion(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	ds.cache.SetAPIRequest(versionReq, version)
+	ds.versionCache.Set(versionCacheKey, version)
 
 	return version, nil
 }
--- a/pkg/zabbixapi/zabbix_api.go
+++ b/pkg/zabbixapi/zabbix_api.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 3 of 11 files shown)


JvstTeo:
I've removed the checklist from the PR description. It was originally in my markdown draft as a personal task tracker, but it's redundant here.

JvstTeo:
Ah, my bad — I hit "Update branch" a couple of times without thinking it
through, which merges main via merge commits instead of rebasing. That
dragged the original unsigned commit back into the history (hence the
"commits could not be verified" report). Caught it on a re-read, rebased
cleanly onto current main instead, and force-pushed a single signed commit.
Should be verified now — thanks for the automated check catching it.

pkg/datasource/resource_handler.go
```diff
@@ -69,7 +70,18 @@ func (ds *ZabbixDatasource) ZabbixAPIHandler(rw http.ResponseWriter, req *http.R
...
+	// QueryData (see datasource.go), so a slow/overloaded Zabbix can't hang
+	// this call indefinitely and trip Grafana's own outer gateway timeout.
+	queryTimeout := dsInstance.Settings.QueryTimeout
+	if queryTimeout <= 0 {
+		queryTimeout = 60 * time.Second
+	}
+	resourceCtx, cancel := context.WithTimeout(req.Context(), queryTimeout)
+	defer cancel()
+
+	result, err := dsInstance.ZabbixAPIQuery(resourceCtx, apiReq)
```
akkikumar72, line 86: The new frontend retry can never fire, because this handler reports every
Zabbix API failure as 500 and the retry only matches 502/503/504.
I pointed a datasource at a stub that always answers 503 and called the
resource endpoint on this branch:
```
HTTP status seen by the frontend: 500
body: {"error":"Internal Server Error","message":"request failed, status: 503 Service Unavailable"}
```
The upstream status survives only inside the message string, so
`RETRYABLE_HTTP_STATUSES.includes(500)` is false and `fetchWithRetry` falls
straight through to `throw error`. The unit tests pass because they inject
`{status: 503}` directly, which skips the handler that flattens it.
Propagating the downstream status fixes it:
```go
result, err := dsInstance.ZabbixAPIQuery(resourceCtx, apiReq)
if err != nil {
	ds.logger.Error("Zabbix API request error", "error", err)
	status := http.StatusInternalServerError
	if backend.IsDownstreamError(err) {
		status = http.StatusBadGateway
	}
	writeError(rw, status, err)
	return
}
```
Worth adding a test that drives the handler with a downstream failure and
asserts the response status, since that is the gap the current tests leave.

pkg/zabbixapi/zabbix_api.go
```diff
@@ -214,30 +218,65 @@ func handleAPIResult(response []byte) (*simplejson.Json, error) {
...
+	if err != nil && isRetryableConnError(err) {
+		// The pooled connection was closed by the server/proxy in the small
+		// window between Go picking it from the idle pool and writing the
+		// request to it (classic keep-alive race, e.g. grafana-zabbix#1295).
+		// Nothing reached the server in that case, so retrying once with a
+		// brand-new connection is always safe, even for non-idempotent
+		// calls like user.login.
+		log.DefaultLogger.Debug("Retrying Zabbix API request after transient connection error", "error", err)
+		body, err = doHTTPRequestOnce(ctx, httpClient, newReq)
+	}
```
akkikumar72, line 261: The retry replays requests that Zabbix already processed, because
`doHTTPRequestOnce` returns the `io.ReadAll(res.Body)` error from line 284
unwrapped, so a truncated response body matches `isRetryableConnError`. By then
the server has sent 200 headers and run the call. A truncated response is also
what an overloaded nginx/php-fpm produces, which is the case this PR is about.
I pointed a datasource at a server that answers 200 with a `Content-Length`
larger than the body it sends, then made a single resource call:
```
--- ONE event.acknowledge call ---
resource endpoint status: 500
DELIVERED AND PROCESSED: apiinfo.version
DELIVERED AND PROCESSED: apiinfo.version
DELIVERED AND PROCESSED: user.login
DELIVERED AND PROCESSED: user.login
```
`user.login` ran twice, which is the method the comment cites as safe. Each run
creates a Zabbix session.
Keeping the body read outside the retry fixes it, so only the round trip is
ever repeated:
```go
func makeHTTPRequest(ctx context.Context, httpClient *http.Client, newReq func() (*http.Request, error)) ([]byte, error) {
	res, err := doRoundTrip(ctx, httpClient, newReq)
	if err != nil && isRetryableConnError(err) {
		log.DefaultLogger.Debug("Retrying Zabbix API request after transient connection error", "error", err)
		res, err = doRoundTrip(ctx, httpClient, newReq)
	}
	if err != nil {
		if backend.IsDownstreamHTTPError(err) {
			return nil, backend.DownstreamError(err)
		}
		return nil, err
	}
	defer func() {
		if err := res.Body.Close(); err != nil {
			log.DefaultLogger.Warn("Error closing response body", "error", err)
		}
	}()
	// Reading the body proves the request was delivered, so a failure here is
	// never a retry candidate.
	return io.ReadAll(res.Body)
}
```

Separately, the round trip itself is not provably safe to replay for POST.
`net/http` already retries the genuine nothing-written case
(`transport.go:842`, when `GetBody != nil`), so what reaches this code is what
`isReplayable()` rejected at `request.go:1534`. Given `script.execute` and
`event.acknowledge` are in `AllowedAPIMethods`, I would skip the retry for
mutating methods too, and reword the comment since "nothing reached the server"
is not what these errors tell us.
A test that fails on the current code: serve a short body with an oversized
`Content-Length` and assert the transport saw exactly one attempt.

pkg/zabbix/methods.go
```diff
@@ -537,7 +537,25 @@ func (ds *Zabbix) GetValueMappings(ctx context.Context) ([]ValueMap, error) {
...
+// (see zabbix.go) checks it on every single API call to pick the right
+// request dialect. Without caching, that turns every logical Zabbix API
+// call the plugin makes into two HTTP requests instead of one, which under
+// real dashboard load (many panels, autorefresh, several users) is enough
+// to exhaust the connection/worker limits of the Zabbix web frontend and
+// makes it start answering with 502/503. Cache the result using the same
+// ZabbixCache already used for host.get/item.get/etc. so the version is
+// only re-fetched once per cache TTL.
 func (ds *Zabbix) GetFullVersion(ctx context.Context) (string, error) {
+	versionReq := &ZabbixAPIRequest{Method: "apiinfo.version"}
```
akkikumar72: Caching the version under a `ZabbixAPIRequest` key puts a `string` into the
same keyspace as the `*simplejson.Json` API responses, and `Request()` reads
that keyspace for every method, not just the cached ones.
The keys differ only by how `params` marshals. `GetFullVersion` leaves `Params`
nil, which hashes `apiinfo.versionnull`, while the frontend sends `{}`, which
hashes `apiinfo.version{}`. Any caller that omits `params` collides, and the
`cachedResult.(*simplejson.Json)` assertion in `zabbix.go:74` then silently
yields an empty object. Reproduced against a real Zabbix 7.0:
```
# on main
$ curl ... -d '{"method":"apiinfo.version"}'
{"error":"Internal Server Error","message":"Invalid request. Invalid parameter \"/params\": an array or object is expected."}
# on this branch, after the version cache is warm
$ curl ... -d '{"method":"apiinfo.version"}'
{"result":{}}
```
Today nothing in the plugin hits that path, since `backendAPIRequest` defaults
`params` to `{}`. But if it ever does, version detection returns empty and the
frontend falls back to `DEFAULT_ZABBIX_VERSION` of `3.0.0`, which silently
picks the wrong request dialect for every query. A dedicated key such as
`c.cache.Set("zabbix_api_version", version)` keeps the two value types apart.

akkikumar72 (commented):
Thanks for the contribution.

I verified the core claim by building the backend from both `main` and this
branch and running the same workload against a real Zabbix 7.0: 2.00 requests
and 2.05 new connections per logical call on `main`, 1.00 and 0.02 on this
branch. Both halves of the fix do what you say they do, and the Go and Jest
suites are green for me too.

Three comments, two must-fix. Both of those are on the hardening extras, not the
version caching or the keep-alive change:
- `makeHTTPRequest` also retries response body read failures, so a truncated
  response replays a request Zabbix already ran. I reproduced `user.login`
  running twice from one logical call.
- the resource handler reports every Zabbix failure as 500, so the new
  502/503/504 frontend retry never fires.

The third is smaller, about the version sharing a cache keyspace with the JSON
responses.

JvstTeo:
@akkikumar72 
Thanks for digging into all three - fixed.

1. Confirmed with the same repro: pointed a stub 503 datasource at
   ZabbixAPIHandler and saw the frontend get `500` with `503` only
   surviving inside the message text, exactly as you found. Added
   `zabbixapi.StatusError` to carry the real status code through the error
   chain, and `resource_handler.go` now uses it instead of hardcoding 500.
   Same repro now returns a real `503`.

2. Fixed - `isRetryableConnError` now only matches errors wrapped in a new
   `preResponseError`, which `doHTTPRequestOnce` only applies to the
   pre-response `ctxhttp.Do` failure path. A body-read failure after a 200
   (even one that's io.EOF-shaped) is never retried now.

3. Fixed - the version is now cached in its own dedicated cache instance,
   fully decoupled from the ZabbixAPIRequest-keyed one used for
   host.get/item.get/etc. No shared keyspace, no collision.

Added a regression test for each: `TestStatusCodeFromError`,
`TestBodyReadFailureAfterOKIsNotRetried`,
`TestVersionCacheDoesNotCollideWithAPIRequestCache`, plus
`TestNon200ResponseCarriesUpstreamStatusCode` for the StatusError itself.

Pushed as a new commit on top instead of amending, so the diff since your
review is easy to see.

akkikumar72:
pkg/zabbixapi/zabbix_api.go

Apart from this CI error, everything else is fine. Please have a look and we can get it merged

CI is failing on `golangci-lint`:

```
pkg/zabbixapi/zabbix_api.go:311:15: QF1011: could omit type error from declaration; it will be inferred from the right-hand side (staticcheck)
    var wrapped error = statusErr
                ^
```

`statusErr` is already `error` (from `fmt.Errorf`), so the explicit type is redundant. Dropping it fixes the lint:

```suggestion
	var wrapped = statusErr
```

JvstTeo:
> Apart from this CI error, everything else is fine. Please have a look and we can get it merged
> 
> CI is failing on `golangci-lint`:
> 
> ```
> pkg/zabbixapi/zabbix_api.go:311:15: QF1011: could omit type error from declaration; it will be inferred from the right-hand side (staticcheck)
>     var wrapped error = statusErr
>                 ^
> ```
> 
> `statusErr` is already `error` (from `fmt.Errorf`), so the explicit type is redundant. Dropping it fixes the lint:
> 
> ```
> 	var wrapped = statusErr
> ```

Ha, funny that lint caught this one. 
Mostly just to avoid an unnecessary variable sitting around - dropped it in favor of two returns in the condition instead.
Reads better this way i guess.

JvstTeo:
Never seen a linter flag something like this before, but glad it did - anyone reading this code later would've gone "wait, why?" 
Had me second-guessing myself for a second there, but it genuinely cracked me up once I actually looked at it.

akkikumar72 (dismissed):
Good job @JvstTeo 🎖️  
Thanks for working on it, we will include this in next release

akkikumar72:
@JvstTeo can you please update this PR with main branch, so we can merge it

JvstTeo:
@akkikumar72 rebased and pushed

akkikumar72 (approved):

JvstTeo:
@akkikumar72 
Can you, please,  tell me, approximately, when the next release with this changes will be?

We're currently using my modified plugin. But... it's not signed. (Obviously)

JvstTeo:
Nvrmd v6.6.0 Is out

Issue #2135, opened 2025-12-10
Add the Suppress option


G0nz0uk:
Hello,

We use the Acknowledge option a lot on the Zabbix plugin, it's great!

We've been asked by a few teams if there is a way to suppress an alert too via Grafana like you can in Zabbix?

<img width="756" height="287" alt="Image" src="https://github.com/user-attachments/assets/e26492a6-d820-4634-ac93-e4278f22f97b" />

Also we can we found out what these disabled options are all about please?

<img width="641" height="93" alt="Image" src="https://github.com/user-attachments/assets/73dd4641-265f-4bf9-8c16-090e6d794876" />

Thanks

zoltanbedi:
It is currently not possible. This is a new feature request.

G0nz0uk:
Ah that's a shame, it would be a game changer for us, thanks for the update though we love this plugin so so much.

Issue #1866, opened 2024-08-27
add suppress alert function


marcilioramos:
1. Currently, there is no native way to suppress Zabbix alerts directly via Grafana. 

2. I would like to be able to suppress Zabbix alerts directly through the Grafana interface. This would allow operators to mute specific alerts without having to access Zabbix separately, making the alert management process more efficient and centralized.

3. An alternative would be to configure suppressions manually in Zabbix itself or via external scripts, but this adds complexity and is not as intuitive as managing them directly in the Grafana interface.

4. This functionality could be integrated into the Grafana alerts panel, with the option to select alerts and define suppression rules based on criteria such as duration, severity or origin. This would make life easier for those who work with active monitoring on multiple platforms.

remote24:
+1

christos-diamantis:
+1

PR #2507, opened 2026-07-29
feat(problems): add unacknowledge, suppress/unsuppress and rank actions

### LINKED ISSUE #2506: Problems panel: support unacknowledge, suppress/unsuppress and rank as cause/symptom actions
The problems panel's "Update problem" dialog currently supports only a subset of the problem update actions the Zabbix API offers:

- acknowledge
- add message
- change severity
- close problem

Zabbix has since added more actions to the `event.acknowledge` method, which are not available from the panel

| Action | API bit | Available since |
| --- | --- | --- |
| Unacknowledge | 16 | Zabbix 5.0 |
| Suppress (indefinitely / until a time ) | 32 + `suppress_until` | Zabbix 6.2 |
| Unsuppress | 64 | Zabbix 6.2. |
| Rank as cause | 128 | Zabbix 6.4 |
| Rank as symptom | 256 + `cause_eventid` | Zabbix 6.4 |

This means common operations (e.g. removing a wrong acknowledgement, temporarily suppressing a noisy problem, or organising cause/symptom relationships) require leaving Grafana and using the Zabbix frontend.

### Expected behavior

- The update dialog offers these actions, matching the Zabbix frontend:
  - Actions should be **state-aware**: acknowledge only for unacknowledged
    problems, unacknowledge only for acknowledged ones, unsuppress only for
    suppressed ones, and rank actions depending on whether the problem is
    already a cause or a symptom.
  - Actions should be **version-gated**: only offered when the connected
    Zabbix server supports them, so nothing breaks on older servers.
- Suppression supports the same time options as Zabbix: indefinitely, for a
  duration, or until a specific date/time (with input validation).
- Ranking a problem as symptom lets the user pick the cause from the problems
  already shown in the panel instead of typing a raw event ID.
- All actions keep recording the acting Grafana user in the event's message
  history ("<user> (Grafana): ..."), as acknowledge does today.

This issue extends #1866 #2135 #2138

christos-diamantis:
Closes #2506 #1866 #2135 #2138 #2548

Had this PR for a while in my backlog. @zoltanbedi did a draft to fix this on #2138 but it was left behind.
It is a very important functionality for daily operations using Grafana plugin and I proceeded with building it.

## Overall 

Extends the Problems panel "Update problem" dialog with the full set of
`event.acknowledge` actions, alongside the existing acknowledge / message /
severity / close:

- **Unacknowledge** (Zabbix 5.0+)
- **Suppress** - indefinitely, for a duration (`30m`, `1h`, `2d`, …), or
  until a specific date (`YYYY-MM-DD HH:mm`), with live validation: invalid
  formats and non-future dates show an error and block submission (Zabbix 6.2+,
  `suppress_until` param)
- **Unsuppress** (Zabbix 6.2+)
- **Rank as cause** (Zabbix 6.4+)
- **Rank as symptom** - the cause is picked from a dropdown of the other
  cause problems currently shown in the panel, instead of typing an event ID
  (Zabbix 6.4+, `cause_eventid` param)

## How it works

- **Version gating**: `queryProblems` stamps `actionCapabilities` on each
  problem DTO (new `supportsUnacknowledge()` / `supportsProblemSuppression()`
  helpers next to the existing 6.4 cause/symptom check), so unsupported
  actions are simply never offered.
- **State gating**: a shared `getAckModalActionProps()` helper derives action
  visibility from the problem itself (`acknowledged`, `suppressed`,
  `cause_eventid`) - acknowledged problems offer *Unacknowledge* instead of
  *Acknowledge*, suppressed ones offer *Unsuppress*, symptoms offer *Rank as
  cause*, etc. Both the table and list layouts use the same helper.
- **Provenance preserved**: every action still sends the "add message" flag
  with the `<user> (Grafana): ...` prefix, so the Zabbix event history keeps
  recording which Grafana user performed each action.
- `acknowledgeEvent()` passes the new `suppress_until` / `cause_eventid`
  params. `event.acknowledge` was already allow-listed in the Go backend, so
  no backend changes were needed.

## Bug fixes in passing

- `ModalController` re-wrapped the modal component with `provideTheme()` on
  every render, remounting the modal (and dropping its state, sometimes
  dismissing it) whenever the panel re-rendered - e.g. on a dashboard refresh
  while the dialog was open. The component is now wrapped once when shown.
- Fixed the stale `acknowledgeEvent` type signatures in `zabbix.ts` /
  `ZabbixConnector` (they had dropped the `action` / `severity` args).
- The panel action buttons now have `aria-label`s.

## Testing

- **Unit** (368 passing): AckModal bitmask building, validation and mutual
  exclusion, state/version gating, cause-dropdown filtering, connector params
  incl. `suppress_until: 0` for indefinite suppression, rank-bit decoding in
  the acknowledges list.
- **E2E** (Playwright vs a real Zabbix 7.0 backend, 20 passing): every action
  exercised through the UI with the wire payload and Zabbix's response
  asserted. Each mutation paired with its inverse so the seeded fixture is
  restored (verified byte-identical after runs). Includes a new provisioned
  dashboard with "hosts in maintenance" shown, needed to reach *Unsuppress*
  since suppressed problems are hidden from the default panel.

## Notes

- Suppressed-by-maintenance problems also expose *Unsuppress* (the panel
  can't currently distinguish manual from maintenance suppression). Zabbix
  ignores manual unsuppression for maintenance windows.

## Screenshots

image

image

image

image


G0nz0uk:
Amazing!  I really hope this gets approved soon, this would really help our 3rd party Helpdesk who use this plugin heavily.

Issue #2464, opened 2026-06-30
Tag filter Operators


gdoring:
**Is your feature request related to a problem? Please describe.**
Currently we have a lot of problems issued by our servers related to disks. However, I want to do something similar to what zabbix does now on their dashboards, which is EXCLUDE a specific tag from showing on problems.

**Describe the solution you'd like**
Allow us to change the operator on the tags filter (like !tag:value)

**Describe alternatives you've considered**
Based on the source code we dont have any alternatives right now except use Zabbix Dashboards, which are ugly.

**Additional context**

<img width="645" height="378" alt="Image" src="https://github.com/user-attachments/assets/623f52bc-0ee5-4796-80f8-f9afef662bd5" />

<img width="900" height="409" alt="Image" src="https://github.com/user-attachments/assets/0d1efc7a-0579-4002-b667-bc64671afc9d" />

itsmylife:
Fixed in https://github.com/grafana/grafana-zabbix/pull/2510

Issue #1421, opened 2022-04-07
RegEXP in tags on problem panel.


serrrios:
Hello!
Need support for regexp in tags in the problem panel to avoid unnecessary problems.

![image](https://user-images.githubusercontent.com/61572759/162169002-d8e5425d-3f17-4447-9960-4ee9afac0ada.png)

gdeleon23:
Hi,
I had a lot of headaches with the hostgroups regex.
Can you try with?

/((?!NOALERT).)*/

Mi regex to exclude "build" and show 24x7 is:

/^((?!BUILD).)*24x7/

Hope this can help you.

alexanderzobnin:
Hi! Do you use problems view in Zabbix? Ho do you use tags there? It's tricky to implement regex filter for tags since it's not supported by Zabbix API. It supports `like` operator which can make filtering more flexible, so maybe I can add it at least.

TheMonkeyBrain:
I am facing the same problem.

@alexanderzobnin 
In zabbix problems view to not show the tag "component:updates" we filter the tags like this:
![zabbix_problem_tags](https://user-images.githubusercontent.com/14860906/185051946-9cb736c8-76f9-453e-b530-4d1d582b7e52.png)

Zabbix v6.0.6
Grafana v9.0.1
Zabbix plugin v4.2.9

I hope this helps

TheMonkeyBrain:
Another solution could be the use of the dashboard settings variables with the regexfilter.
But it doesn't look like i can get the tags through that.

lomobot:
@alexanderzobnin Mr.Zobnin, our team needs this feature vey much

mrgolangv:
Me too

klavik:
@alexanderzobnin It might be very usefull for my team too.

timnorm:
@alexanderzobnin very necessary functionality!

demanuPL:
Any news about this feature? T clould be good to exclude problems by filtering tags by regex

PR #2510, opened 2026-07-30
feat(problems): support all Zabbix tag-filter operators in problems queries

### LINKED ISSUE #2509: Problems panel: support tag filter operators like Zabbix
When the query type is **Problems**, the tag filter only supports a free-text `tag1:value1, tag2:value2` field, and every tag is queried with a fixed operator (Equal). Since Zabbix 5.4, `problem.get`/`event.get` support the full set of tag-filter operators - Contains, Equals, Does not contain, Does not equal, Exists, Does not exist — the same set the Zabbix frontend offers in Monitoring → Problems, but none of these can be expressed in the plugin. Common filters like "show problems where tag `service` exists" or "exclude `environment: staging`" are impossible today.

Replace the free-text tags field in the Problems query editor with a structured tag filter editor matching the Zabbix Problems view: one row per filter with a tag name, an operator picker, and a value, plus Add/Remove and the existing AND/OR-OR evaluation toggle.

* Version-gated: extended operators are only offered on Zabbix ≥ 5.4 (where the API accepts them); older versions keep Equals and Contains. Pre-7.0 operator labels ("Not like", "Not equal", "Not exists") are used on 5.4-6.x, matching what the host tag filter already does.
* Backward compatible: existing dashboards with the legacy text filter must keep returning identical results without being touched - migrated on the fly at query time to structured filters with the Equal operator the old code hardcoded.
* The tag name field should suggest tags from the problems the panel already fetched (no extra API calls, to avoid additional load on the Zabbix frontend).
* The Problems panel's click-to-filter (adding/removing a tag filter by clicking a tag chip) must work with the new model too.

* Extending the free-text syntax with operator tokens (e.g. `tag!=value`, `tag=~value`) - cryptic, error-prone to parse, and diverges from how Zabbix itself presents these filters.
* Adding a single global operator dropdown next to the existing text field - cannot express per-tag operators, which the Zabbix API supports.

**Additional context**
Zabbix API reference for `problem.get` tags: operators 0–5 available since 5.4 (5.2 and older only accept 0 Like / 1 Equal). The plugin already ships a structured tag filter with these operators for **host** tags in the Metrics editor, so this brings the Problems editor to parity.

christos-diamantis:
Closes #2509 #2464

## Overall

Adds full Zabbix tag-filter operator support to **Problems** queries. The free-text `tag1:value1` field is replaced by a structured tag filter editor matching the Zabbix Problems view: per-filter tag name, operator, and value, with Add/Remove rows and the AND/OR-OR evaluation toggle.

### Operators & version gating

* All six operators supported: **Exists, Equals, Contains, Does not exist, Does not equal, Does not contain**.
* Extended operators (2-5) were added to `problem.get`/`event.get` in **Zabbix 5.4.0** (5.2 only accepts Like/Equal), so the editor gates them behind a new `supportsProblemTagOperators()` check (≥ 5.4). On 5.4-6.x the negated operators use the pre-7.0 labels ("Not like", "Not equal", "Not exists"), reusing the existing host-tag label helper.
* Exists/Does not exist hide the value input and omit `value` in the API param, as Zabbix ignores it.

### Backward compatibility (no broken dashboards)

* New query field `problemTags` + query schema bump to **13**. Legacy `tags.filter` text is migrated to structured filters with **operator 1 (Equal)** - exactly what the old code hardcoded - so results are identical.
* Old dashboards keep working **without being saved**: targets are migrated on a clone at query time; the stored JSON is only rewritten when the user edits the query.
* The legacy text path remains as a fallback in `queryProblems`, and template variables that expand to the legacy `tag:value[, ...]` format are still split correctly after interpolation.
* The Problems panel click-to-filter (tag chips) now writes whichever model the target uses.
* Regression-pinned: a unit test asserts a migrated legacy target produces the byte-identical `problem.get` `tags` param the old implementation sent, and an e2e test replays a real pre-schema-13 dashboard against a live Zabbix.

### Tag name suggestions at zero API cost

The tag combobox suggests tags extracted from the problems the panel already fetched (`data` prop; `problem.get` already runs with `selectTags: 'extend'`) - no additional API requests, so no extra load on the Zabbix frontend even on large dashboards. Template variables stay at the top of the list and free-typed tags are still accepted.

## Testing

* **Unit (Jest, 328 passing):** editor component (operator gating per version, labels, add/remove/edit, value hiding), query editor wiring, schema-13 migration cases, tag param builder, version gate.
* **E2E (Playwright vs real Zabbix, 14 passing):** new `problemsTagFilter.spec.ts` with seeded tagged problems -
  * a legacy dashboard (`schema: 12`, text filter) sends `[{ tag, value, operator: 1 }]` and renders only the matching problem;
  * a structured dashboard using **Exists** sends `[{ tag, operator: 4 }]`, is accepted by Zabbix, and filters at the source.
* Docs (`query-editor.md`) included.

## Notes for reviewers

* Operator enums in `QueryEditor/types.ts` were generalized (`TagOperatorValue` etc.) since host and problem tags share the same protocol values. the old `HostTag*` names are re-exported as aliases, so no call sites changed.
* If a dashboard built on ≥ 5.4 with extended operators is pointed at an older Zabbix, the panel surfaces Zabbix's own "invalid operator" API error rather than silently dropping the filter and showing wrongly-filtered data.

### Old tag selector
image

### New tag selector
image
image
image
image


.changeset/six-hounds-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'grafana-zabbix': minor
+---
+
+Problems query: support all Zabbix tag-filter operators (Exists, Equals, Contains, Does not exist, Does not equal, Does not contain). The free-text tags field is replaced by a structured tag filter editor matching the Zabbix Problems view — each filter has a tag name, an operator picker, and a value. Extended operators are offered only on Zabbix 5.4+, where `problem.get`/`event.get` accept them; older versions keep Equals and Contains. Existing dashboards are migrated automatically: legacy `tag1:value1` text filters are converted to structured filters with the Equals operator they were previously queried with.
```
itsmylife: Could you please shorten the changelog?

itsmylife (changes requested):
This looks good to me. Here are a few nits

Please use `Closes #2509, closes #2464`. Leave [#1421](https://github.com/grafana/grafana-zabbix/issues/1421) open (regexp is not in the Zabbix tag API). And I don't think this PR is resolving that issue.

.changeset/six-hounds-brake.md
christos-diamantis: shortened on the latest commit

christos-diamantis:
Indeed, #1421 was for regex but was also discussing all the tag operators. I removed it from the PR body, so it is not closed. 
Fix the merge conflict and also shortened the changelog message. should be ok for merging

itsmylife (approved):

Issue #2329, opened 2026-03-04
Bug: `percentile()` fails to parse `$__range_series` in Grafana-Zabbix plugin


Miccia94:
**Description**

Using the `percentile()` transform function with the documented parameter `$__range_series` causes the query to fail with a duration parsing error.

The backend receives the literal string `"$__range_series"` instead of an expanded value and attempts to parse it as a Go duration, which results in a failure.

**Error**

```
failed to parse function param: time: invalid duration "$__range_series"
```

**Environment**

* Grafana: 12.4.0
* grafana-zabbix plugin: 6.2.1
* Datasource: Zabbix
* Query type: `time_series`

**Steps to reproduce**

1. Create a panel using the Zabbix datasource.
2. Query any item returning time-series data (e.g. interface traffic).
3. Add the transform function:

```
percentile($__range_series, 95)
```

4. Execute the query.

**Expected behavior**

The percentile should be calculated over the entire series range as described in the documentation.

**Actual behavior**

The query fails with:

```
failed to parse function param: time: invalid duration "$__range_series"
```

Query Inspector shows that the macro is not expanded before being parsed.

**Query Inspector (request excerpt)**

```json
{
  "functions": [
    {
      "name": "percentile",
      "params": [
        "$__range_series",
        95
      ],
      "text": "percentile($__range_series, 95)"
    }
  ]
}
```

This indicates the backend receives the literal value `"$__range_series"` instead of an expanded value.

**Additional context**

* Explicit durations such as `1h`, `7d`, or `30d` work correctly.
* The issue appears specific to macro expansion of `$__range_series` when used as the interval parameter of `percentile()`.

This suggests the macro is not expanded before the Zabbix plugin attempts to parse the interval parameter.

herraa1:
The bug is again reproducible with:

Environment

Grafana: 13.1.0
grafana-zabbix plugin: 6.4.0
Datasource: Zabbix
Query type: time_series, gauge

Identical as initial report.

PR #2511, opened 2026-07-30
fix(metrics): percentile() fails to parse $__range_series and $__range

### LINKED ISSUE #2469: Bug: percentile() fails again to parse $__range_series in grafana-zabbix plugin
Same as #2329, but reproducible again with the latest available alexanderzobnin-zabbix-app @ 6.4.0

Description

Using the percentile() transform function with the documented parameter $__range_series causes the query to fail with a duration parsing error.

The backend receives the literal string "$__range_series" instead of an expanded value and attempts to parse it as a Go duration, which results in a failure.

Error

failed to parse function param: time: invalid duration "$__range_series"
Environment

Grafana: 13.1.0
grafana-zabbix plugin: 6.4.0
Datasource: Zabbix
Query type: time_series
Steps to reproduce

Create a panel using the Zabbix datasource.
Query any item returning time-series data (e.g. interface traffic).
Add the transform function:
percentile($__range_series, 95)
Execute the query.
Expected behavior

The percentile should be calculated over the entire series range as described in the documentation.

Actual behavior

The query fails with:

failed to parse function param: time: invalid duration "$__range_series"
Query Inspector shows that the macro is not expanded before being parsed.

Query Inspector (request excerpt)

{
  "functions": [
    {
      "name": "percentile",
      "params": [
        "$__range_series",
        95
      ],
      "text": "percentile($__range_series, 95)"
    }
  ]
}
This indicates the backend receives the literal value "$__range_series" instead of an expanded value.

Additional context

Explicit durations such as 1h, 7d, or 30d work correctly.
The issue appears specific to macro expansion of built-in variables like $__range_series and others described at https://grafana.com/docs/plugins/alexanderzobnin-zabbix-app/latest/functions/ when used as the interval parameter of percentile().
This suggests the macro is not expanded before the Zabbix plugin attempts to parse the interval parameter.

christos-diamantis:
Closes #2469 

## Overall

Fixes `percentile($__range_series, 95)` (and any other query function using a range macro) failing with:

`failed to parse function param: time: invalid duration "$__range_series"`

## Root cause
The `DataSourceWithBackend` migration (#2123) moved variable interpolation to a single `interpolateVariablesInQueries(requestTargets, request.scopedVars)` call at the top of `query()`. However, the range scoped vars (`$__range`, `$__range_ms`, `$__range_s`, `$__range_series`) were only merged into `request.scopedVars` later - inside `frontendQuery()`/`dbConnectionQuery()` - after backend targets had already been interpolated and dispatched.
Before the migration, the old `backendQuery()` merged `getRangeScopedVars(request.range)` into scoped vars *before* calling `replaceTargetVariables()`, so `$__range_series` was correctly expanded to the `range_series` sentinel that the Go backend special-cases. After the migration, the backend received the literal macro string and tried to parse it as a Go duration.
The earlier fix attempt (#2368) was closed without being merged, which is why #2329 was reported again as #2469.

## Changes
- **Frontend (root cause):** merge `utils.getRangeScopedVars(request.range)` into the scoped vars before `interpolateVariablesInQueries()` in `query()`, restoring expansion of all range macros in function params for backend, direct-DB, frontend, and annotation targets.
- **Backend (safety net):** add `isRangeSeriesVariable()` that accepts the expanded `range_series` sentinel as well as the literal `$__range_series` / `${__range_series}` forms in `groupBy`, `percentile`, and `percentileAgg`. This covers alerting, where queries reach the backend directly and frontend interpolation never runs.
- **Tests:** Go tests covering both the sentinel and raw-macro forms; a frontend regression test asserting `query()` passes `__range_series` / `__range` / `__range_s` / `__range_ms` into interpolation. The test fixture's time range now uses real `dateTime` objects since `getRangeScopedVars()` needs `.diff()`.

## How to reproduce / verify
1. Create a panel with the Zabbix datasource, query any numeric item.
2. Add the function `percentile($__range_series, 95)`.
3. Before: query fails with the duration parse error. After: the percentile is calculated over the whole series range.


.changeset/wet-chairs-wear.md
```diff
@@ -0,0 +1,5 @@
+---
+'grafana-zabbix': patch
+---
+
+Fix query functions failing with `time: invalid duration "$__range_series"` when using range macros as the interval parameter (e.g. `percentile($__range_series, 95)`). The DataSourceWithBackend migration moved variable interpolation to the start of the query flow, but the range scoped vars (`$__range`, `$__range_ms`, `$__range_s`, `$__range_series`) were only added afterwards, so backend queries received the macro unexpanded. Range vars are now in scope before interpolation, and the backend additionally accepts the literal `$__range_series` macro so alerting queries — which never pass through frontend interpolation — also work.
```
itsmylife: Could you please make this one liner? It's a too long changelog entry

itsmylife (changes requested):
Looks good to me. I requested a small thing only.

.changeset/wet-chairs-wear.md
christos-diamantis: @itsmylife i did change it to a really shorter version. I was using claude for generating a good changeset, but indeed it had too much info!

itsmylife (approved):

PR #2513, opened 2026-07-30
feat(problems): add optional Host IP field to problems panel

### LINKED ISSUE #2512: Problems panel: Add host IP option

The Problems panel can show the host name, technical name, host groups,
and proxy, but there is no way to see the host's IP address. When
triaging problems, operators often need the IP to reach the affected
machine (SSH, ping, inventory lookup) and currently have to open Zabbix
and navigate to the host configuration to find it.

An optional **Host IP** field on the Problems panel (both table and
list layouts), turned off by default, displayed after Host name and
Host technical name. The IP should come from the host's interfaces:

- if the host has more than one interface, show all IPs comma-separated
- if the host has no interfaces, show an empty string
- if an interface is configured to connect via DNS instead of IP, it
  should contribute nothing (a DNS-only host shows an empty string)

Since host interfaces are not part of the data currently fetched for
problems, the extra API call should only happen when the feature is
enabled, so dashboards that don't use it pay no extra cost.

- Using trigger.get `selectHosts` - not possible, the Zabbix API does
  not return interfaces through that selector.
- A separate `hostinterface.get` call - not in the plugin's API method
  allowlist; `host.get` with `selectInterfaces` achieves the same with
  no backend changes.

christos-diamantis:
Closes #2512 

## Overall

Adds an optional **Host IP** field to the Problems panel, off by default, shown after **Host name** and **Host technical name** in both the table and list layouts.

The IP is resolved from the host's interfaces:

- multiple IP-based interfaces → all IPs, comma-separated
- no interfaces → empty string
- interfaces connecting via DNS (`useip = 0`) are ignored → DNS-only hosts show an empty string

## How

Follows the existing **Host proxy** pattern - two toggles working together:

1. **"Host IP" query option** (query editor → Options) gates the data fetch. This is required because the datasource never sees panel options.
2. **"Host IP" field toggle** (panel options → Fields) shows the column.

### Keeping the API overhead minimal

- Zabbix's `trigger.get`/`event.get` cannot return interfaces via `selectHosts`, and `hostinterface.get` is not in the API allowlist. Instead, one `host.get` call with `output: ['hostid']` and `selectInterfaces: ['ip', 'useip']` is used - already allowlisted, so **no Go/backend changes**.
- The call is made **only when the query option is enabled**, **only for the hosts present in the current result set**, and is skipped entirely when no problems/hosts.
- `getHostInterfaces` is registered in the caching proxy, so repeated refreshes with a stable problem set hit the cache.

## Changes

- `zabbixAPIConnector.getHostInterfaces()` - minimal `host.get` + `selectInterfaces` call
- `problemsHandler.addTriggerHostIps()` - maps interfaces to problems and applies the IP/DNS/empty rules
- `datasource.queryProblems()` - gated enrichment step after the proxy step
- Query option + editor switch, panel option + Fields switch, table column, list-layout span, docs entry in `query-editor.md`

## Tests

- `problemsHandler.test.ts` - 8 cases: single/multiple IPs, no interfaces, DNS-only, mixed DNS+IP+empty, host missing from response, hostless problems, multi-host mapping
- `datasource.test.ts` - `getHostInterfaces` called only when the option is on and hosts exist; IP lands on the problem
- `zabbixAPIConnector.test.ts` - exact minimal `host.get` payload
- `Problems.test.tsx` - column hidden by default, shown when enabled, comma-separated rendering, added to sortable-column tests
- `AlertCard.test.tsx` - new suite (list layout previously untested): hidden by default, rendered after host name, comma-separated IPs, nothing rendered when empty


itsmylife (changes requested):
Thanks for the PR. It looks good. I request a nit only.

christos-diamantis:
fixed the merge conflicts. @itsmylife i do not see any requested change tbh

itsmylife:
> fixed the merge conflicts. @itsmylife i do not see any requested change tbh

I forgot to submit that nit but anyways. It look good. I'm gonna approve. Thanks for the PR and swift reply!

itsmylife (dismissed):

src/datasource/specs/datasource.test.ts
```diff
@@ -102,6 +102,73 @@ describe('ZabbixDatasource', () => {
...
+      const ds = new ZabbixDatasource(instanceSettings);
+      const zabbixMock = buildZabbixMock([{ name: 'Hostless problem', suppressed: '0', hosts: [] }]);
+      ds.zabbix = zabbixMock as any;
+
+      await ds.queryProblems(buildProblemsTarget({ hostIp: true }), [0, 100], {});
+
+      expect(zabbixMock.getHostInterfaces).not.toHaveBeenCalled();
+    });
+  });
+
```
itsmylife: Could you please delete 2 spaces in this line to make linter happy

christos-diamantis: just fixed on the latest commit. it happened on the previous merge conflict resolution.

christos-diamantis:
@itsmylife fixed the linting issue. it was coming from a merge conflict resolution earlier. i did not run all the tests afterwards and did not see the lint error. the CI should be happy now

itsmylife (approved):

PR #2525, opened 2026-08-17
feat(dataplane): add support for grafana data plane

### LINKED ISSUE #2524: Grafana data plane support

Query responses from the plugin carry no [data plane](https://grafana.github.io/dataplane/contract/) type declaration: no frame sets `meta.type` / `meta.typeVersion`. Consumers therefore have to infer the shape of the response, which has a few concrete consequences:

- **Server-side expressions and alerting guess the shape.** Grafana's `expr` package only takes the data plane path when the first frame declares a type and a non-zero version; otherwise it falls back to inferring a series from the frame. Anything the inference gets wrong surfaces to users as an error unrelated to what they were doing.
- **SQL expressions need the declaration.** Per [Requirements to support SQL expressions](https://grafana.com/developers/plugin-tools/how-to-guides/data-source-plugins/sql-requirements), the data plane contract is mandatory for labeled data, so Zabbix queries can't be used there.
- **The series name isn't where the contract says it is.** The contract states that "the series name comes from the Value Field's Name property", while the plugin names every value field `Value` and carries the name only in `FieldConfig.DisplayNameFromDS`. In server-side expressions this means every series is identified as `Value`.
- **The response shape is inconsistent.** The backend emits one frame per series (already structurally `timeseries-multi`), but the frontend then collapses the response into a single "wide" frame whenever the first two timestamps of each frame happen to match - merging frames across different `refId`s in the process. So the same query returns a different shape depending on the number of series and on whether that two-point check passes.
- **The time field isn't guaranteed to be sorted.** Only trend data is sorted; `history.get` is sorted by `clock`, which has second precision, so points within the same second can be returned out of order. The contract requires an ascending time field.

Make every query response data plane compliant:

- Declare `timeseries-multi` (version 0.1) on the numeric series frames produced by Metrics and Item ID queries, and `timeseries-wide` on the frames that are natively wide (SLI/SLA, Service status), plus the trigger-count frame.
- Move the series name onto the value field, keeping `displayNameFromDS` so displayed names, legends and existing dashboard overrides are unchanged.
- Always sort series by time, not only trend data.
- Remove the frontend wide conversion so the declared type always matches what is returned, and let the contract's multi format do what it exists for: series that may not share a time index.
- Cover it with a test that validates real responses through `grafana/dataplane/sdata`, the same reader Grafana uses.

- **Declare the types but keep the `Value` field name.** Non-breaking, and `sdata` accepts it, but it's semantically wrong and makes it worse in practice: the alert rule editor and expression previews would identify every series as `Value`, where today they show the Zabbix series name (legacy expression handling renames the value field to the frame name). Since the two changes are only equivalent when done together, they belong in the same release.
- **Keep the wide conversion and declare `timeseries-wide` instead.** This would mean asserting a shared time index on the strength of a check that compares only the first two timestamps of each frame, and it would keep merging series from different queries into one frame, which the contract treats as one type per response.
- **Gate the renaming behind a datasource option for one release.** Doubles the shapes to support and to test, and leaves the alerting-side naming regression in place for whoever doesn't flip the toggle.

**Additional context**

The value fields' labels (`host`, `item`, `item_key`) stay exactly as they are, so alert rule instance labels, silences and notification templates are unaffected - both the legacy and data plane paths in `expr` take their labels from the value field.

christos-diamantis:
Closes #2524

Makes every query response follow the [data plane contract](https://grafana.github.io/dataplane/contract/timeseries).

## What changed

**Backend**

- `seriesToDataFrame` declares `timeseries-multi` / typeVersion `0.1` on each series frame, and names the value field after the series (`srv01: CPU load`) instead of `Value`, falling back to `Value` for an unnamed series. `DisplayNameFromDS` is kept, which is what keeps legends and display-name-based overrides identical.
- Series are now always sorted by time in `applyDataProcessing`, not only when trends are used: `history.get` is sorted by `clock` (second precision), so points sharing a second could arrive out of order. `TimeSeries.Sort()` short-circuits on already-sorted input, so this is free in the common case.

**Frontend**

- Removed the wide conversion (`convertToWide`, `isConvertibleToWide`, `getLongestFrame`) and all four call sites. Every series is returned as its own frame, so the shape always matches the declared type. Previously the response collapsed into a single frame whenever the first two timestamps of each frame matched, merging frames across `refId`s.
- SLI/SLA and Service status frames declare `timeseries-wide` (they were already natively wide: one shared time field, one named value field per service). The trigger-count frame declares `timeseries-multi` and its value field is renamed to `Count <refId>` with a `displayNameFromDS` so the displayed name stays exactly what the frame name previously resolved to.
- Alias functions (`setAlias`, `setAliasByRegex`, `replaceAlias`) now update `field.name` alongside `displayNameFromDS`, so an alias moves the series name too.
- Fixed `alignFrames` padding only the first value field, which left frames holding more than one series (SLI) at unequal field lengths and, because `MutableDataFrame` derives its row count from a cached reference to the first field's values, left `frame.length` stale as well.

Text mode and the table-shaped frames (problems, macros, history-as-table, triggers table) are intentionally left undeclared: the time series kind only permits numeric and bool value fields, and `table` has no versioned contract, so a non-zero typeVersion there would make `sdata` reject the response and log a fallback warning.

## Breaking changes

- **Value fields are named after the series instead of `Value`.** Legends, panel titles and overrides that match on the displayed name are unaffected. Transformations and overrides referencing the literal field name `Value` (Organize fields, Filter fields by name, Table panel column headers, CSV export headers) need updating to the series name. This only affects single-series responses in practice, since the removed wide conversion already renamed value fields to series names for multi-frame responses.
- **Responses are no longer collapsed into a single wide frame.** Time series, Stat, Gauge and Table panels handle multiple frames directly; panels that need a single frame (Bar chart, Pie chart, XY chart) now require a `Join by field` transformation on the time field, and Table panels show a frame selector instead of merged columns.

Alert rules are not affected. Both the legacy and data plane paths in Grafana's `expr` package take their labels from the value field, and `host` / `item` / `item_key` are unchanged, so alert instance labels, silences and notification templates keep matching. The series name that expressions see is also unchanged: legacy handling renamed the value field to the frame name, which is the same string this PR writes into the field name.

## Testing

- New `pkg/datasource/dataplane_test.go` validates real responses through `grafana/dataplane/sdata` with **strict** validation (`GetCollection(true)`) - the same reader Grafana's server-side expressions use, and stricter than what alerting itself applies. Added as a direct dependency.
- New frontend tests for the type declarations and value field naming, and for `alignFrames` keeping every field the same length.
- `go vet` clean, Go tests pass, `yarn typecheck` clean, `yarn lint` at 0 errors, 294 jest tests pass.
- Verified end to end against Grafana 12.3.1 with the plugin's devenv. A query with a `reduce` expression returns the correct value with labels intact, and Grafana logs the path it took:

  ```
  logger=expr level=debug msg="Handling SSE data source query through dataplane" datatype=timeseries-multi
  logger=expr level=debug msg="Data source queried" datasourceType=alexanderzobnin-zabbix-datasource queryRefId=A responseType=dataplane-timeseries-multi
  ```

  No `"Dataplane data detected but falling back to old processing due to error"` warning, which is what Grafana logs when `sdata` rejects the frames and reverts to the legacy path.


christos-diamantis:
Any update?

PR #2527, opened 2026-08-17
feat: add ClickHouse support for Direct DB Connection (Zabbix 8.0 history storage)

### LINKED ISSUE #2526: Support ClickHouse as Direct DB Connection backend (Zabbix 8.0 history storage)
## Feature request

Zabbix 8.0 adds ClickHouse as a native history storage backend
([What's new](https://www.zabbix.com/documentation/8.0/en/manual/whatsnew),
[ClickHouse setup](https://www.zabbix.com/documentation/8.0/en/manual/appendix/install/clickhouse_setup)).
The plugin's Direct DB Connection currently supports only MySQL, PostgreSQL, and InfluxDB, so
users moving history to ClickHouse lose the direct-DB fast path and fall back to the Zabbix API
for all history queries.

## Why this matters

- Zabbix 8.0 is releasing soon and ClickHouse is its recommended high-volume history backend.
- Direct DB Connection is most valuable exactly where ClickHouse is used: large installations
  with wide time ranges.

## Details relevant to the implementation

The official schema (`database/clickhouse/` in the zabbix repo) differs from the SQL backends:

- Per-value-type tables: `history`, `history_uint`, `history_str`, `history_text`, `history_log`, `history_json`
- Columns are `itemid UInt64`, `clock_ns DateTime64(9)`, `value` - there is **no integer `clock`/`ns` pair**,
  so the existing MySQL/PostgreSQL dialects cannot be reused as-is
- **Trends are not calculated or stored in ClickHouse**; retention is TTL-based. Trends requests
  must be served from history data.
- If a ClickHouse datasource is selected today, `Zabbix.initDBConnector` silently falls through to
  the SQLConnector with the MySQL dialect and produces invalid queries.

## Proposal

Add a dedicated ClickHouse connector targeting the official `grafana-clickhouse-datasource`
plugin, list it in the Direct DB Connection datasource picker, serve `getTrends` from history,
and document the setup (default database must point to the Zabbix history database).

christos-diamantis:
Closes #2526

## What this PR does

Adds ClickHouse as a supported Direct DB Connection backend, so the plugin can query the
Zabbix 8.0 ClickHouse history storage directly through an existing
[`grafana-clickhouse-datasource`](https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/).

## Implementation notes

- **New `ClickHouseConnector`** (`src/datasource/zabbix/connectors/clickhouse/clickhouseConnector.ts`)
  - Builds aggregated history SQL against the official Zabbix 8.0 ClickHouse schema
    (`itemid UInt64`, `clock_ns DateTime64(9)`, `value`): `intDiv(toUnixTimestamp(...))` interval
    bucketing, `toDateTime64` range filters that hit the `(itemid, clock_ns)` primary key, and
    `consolidateBy`-driven aggregation, mirroring the existing SQL connector semantics.
  - Posts to `/api/ds/query` in the ClickHouse plugin's query shape (`editorType: 'sql'`,
    `rawSql`, table format) and pivots the returned long `metric/time/value` frame into per-item
    frames (`field.name = itemid`, ms timestamps) client-side. The `time_series` format used by
    the core SQL datasources was deliberately avoided: the ClickHouse plugin's wide-format
    conversion names value fields differently, which would break itemid matching in
    `dataResponseToTimeSeries`.
- **Trends**: Zabbix does not calculate or store trends in ClickHouse, so `getTrends()` delegates
  to `getHistory()` (same pattern as the InfluxDB connector). Wide time ranges that would normally
  switch to trends are served from history with server-side aggregation, regardless of the trends
  settings.
- **Connector selection**: `Zabbix.initDBConnector` now handles `grafana-clickhouse-datasource`
  explicitly - previously it would silently fall through to the SQLConnector with the MySQL dialect.
- **ConfigEditor**: ClickHouse added to the supported datasource list; an info alert explains the
  history-only behavior and that the ClickHouse datasource's **default database** must point to the
  Zabbix history database (`zabbix` by default), since queries don't prefix table names.
- **Docs**: `configure.md` (new "ClickHouse history storage" section), `query-editor.md`,
  `troubleshooting.md`, `_index.md`.

## Testing

- New `clickhouseConnector.test.ts`: query generation, per-value-type table selection,
  trends→history delegation, `/api/ds/query` payload, long-frame pivot (incl. empty/malformed
  responses).
- `zabbix.test.ts`: connector selection per datasource type (InfluxDB / ClickHouse / SQL fallback).
- Full suite: 21 suites / 303 tests pass; `tsc --noEmit` and ESLint clean.


christos-diamantis:
Any update?

PR #2531, opened 2026-08-18
Chore: Use npm as package manager

itsmylife:
- Switch to npm as package manager
- Update docs
- Remove redundant yarn entries


yesoreyeram (approved):
nice one. Thank you..!

PR #2532, opened 2026-08-19
Chore: Update @grafana/create-plugin and .config folder

itsmylife:
- Update `.config` folder with a fresh plugin build
- add agents files


itsmylife:
@christos-diamantis Nice catch. I'll fix it.

.config/bundler/externals.ts
```diff
@@ -16,6 +16,8 @@ export const externals: ExternalsType = [
   'slate-plain-serializer',
   '@grafana/slate-react',
   'react',
+  'react/jsx-runtime',
```
yesoreyeram, line 19: do we need to update https://github.com/grafana/grafana-zabbix/blob/45b17b398d43e3638cf9fef6ff809f82e590456e/.config/docker-compose-base.yaml#L10 related to this externalisation?

.config/Dockerfile
```diff
@@ -7,8 +7,9 @@ ARG anonymous_auth_enabled=true
 ARG development=false
 ARG TARGETARCH
 
-ARG GO_VERSION=1.25.6
+ARG GO_VERSION=1.21.6
```
yesoreyeram: seems mistakenly downgraded. Let's go with 1.27.1

.config/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG grafana_version=latest@sha256:b92dd301ac40f837e87bfd05428397f9096a0941773ab417e67b399d0381b3ba
+ARG grafana_version=latest
```
yesoreyeram: shouldn't we pin it?

yesoreyeram (changes requested):
left some nit comments

.config/Dockerfile
itsmylife: We should. Reverted

.config/Dockerfile
itsmylife: I'd stick with `1.26.5` as main go.mod is still in that version.

.config/bundler/externals.ts
itsmylife: yeah let's pin that to 13.1.6

yesoreyeram (approved):

PR #2541, opened 2026-08-28
fix(functions): scale() fails when added with its default param

### LINKED ISSUE #2540: scale() function fails with "failed to parse function param" when added with its default factor
### Describe the bug

Adding the `scale()` function to a metrics query immediately breaks the panel with a query error. The function is inserted with its default factor (100) and the backend rejects it. Deleting the value and retyping the same `100` makes the query work.

### Steps to reproduce

1. Open a metrics query in the query editor
2. Add function → Transform → `scale`
3. The panel errors as soon as the query runs with the inserted default

Image

Retyping the factor (any value) fixes it, which is why this is easy to misread as a transient editor glitch.

### Root cause

Type mismatch between how the default param and an edited param are serialized:

- `metricFunctions.ts` defines `scale` with `defaultParams: [100]` - a **number**. `FuncInstance` copies it into `params` as-is, so a freshly added function is saved as `params: [100]`.
- `QueryFunctionsEditor.onFuncParamChange` stores edited values as **strings** (`params: ["100"]`), since the segment editor deals in strings.

On the backend, `applyScale` (`pkg/datasource/functions.go`) is the only numeric function that hard-requires a string param:

```go
pFactor, err := MustString(params[0])   // fails on JSON number 100
...
factor, err := strconv.ParseFloat(pFactor, 64)
```

Every comparable function (`offset`, `removeAboveValue`, `transformNull`, `percentile`, …) uses `MustFloat64`, which accepts both `float64` and string - which is why `offset()` works fine despite having the identical `defaultParams: [100]`.

This also means any dashboard saved with the numeric form keeps failing until the param is manually retyped.

### Expected behavior

`scale()` works when added with its default, like every other numeric function.

### Environment

- Grafana-Zabbix: reproduced on current main
- Affects backend metric processing (`applyScale`)

Fix incoming as a PR: switch `applyScale` to `MustFloat64`, matching the other numeric functions.

christos-diamantis:
Fixes #2540

`scale()` errored with `failed to parse function param: failed to convert value to string: 100` whenever it was added through the editor and ran with its inserted default. Retyping the same value made it work.

The frontend serializes the two cases differently: `defaultParams: [100]` is copied into the query as a JSON **number**, while values typed in the editor are stored as **strings**. `applyScale` was the only numeric function using `MustString` + `strconv.ParseFloat`, so it rejected the numeric form; its siblings (`offset`, `removeAboveValue`, `transformNull`, …) all use `MustFloat64`, which accepts both.

### Changes

- `applyScale` now uses `MustFloat64(params[0])`, matching the other numeric functions. `MustFloat64` already falls back to `strconv.ParseFloat` for strings, so template-substituted values keep working.
- Added `TestApplyScale` covering the numeric (freshly added default), string (typed), and invalid param cases.

### Notes

- Fixing this backend-side (rather than changing `defaultParams` to `['100']`) also repairs dashboards already saved with the numeric param - no retyping needed.
- No frontend changes; the frontend path already handles both types via `Number(param)`.

### DIFF
--- a/.changeset/slimy-trees-occur.md
+++ b/.changeset/slimy-trees-occur.md
@@ -1,0 +1,5 @@
+---
+'grafana-zabbix': patch
+---
+
+fix(functions): accept numeric scale() param; add frontend/backend param-contract tests
--- a/pkg/datasource/functions.go
+++ b/pkg/datasource/functions.go
@@ -214,11 +214,7 @@ func applyPercentile(series timeseries.TimeSeries, params ...interface{}) (times
 }
 
 func applyScale(series timeseries.TimeSeries, params ...interface{}) (timeseries.TimeSeries, error) {
-	pFactor, err := MustString(params[0])
-	if err != nil {
-		return nil, errParsingFunctionParam(err)
-	}
-	factor, err := strconv.ParseFloat(pFactor, 64)
+	factor, err := MustFloat64(params[0])
 	if err != nil {
 		return nil, errParsingFunctionParam(err)
 	}
--- a/pkg/datasource/functions_test.go
+++ b/pkg/datasource/functions_test.go
@@ -1,6 +1,11 @@
 package datasource
 
 import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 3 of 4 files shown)


itsmylife (approved):
LGTM!

PR #2543, opened 2026-09-02
fix(problems): expand trigger description macros on the default query path

adamyeats:
## Summary

- ask Zabbix to expand the trigger comment on `trigger.get`, so the Problems panel Description no longer shows raw `{ITEM.VALUE}` text with default query options
- keep the parameter off when "Item value at problem time" is enabled, so the plugin can still expand each problem's own value
- add unit coverage plus an e2e test and fixture against the seeded Zabbix backend

Since 6.4.1 neither side expanded these macros on the default path. #2404 removed `expandComment` from `trigger.get` so the plugin could expand the comment itself, per problem. #2460 then put that client-side expansion behind the "Item value at problem time" option, which is off by default, and did not restore the server-side flag.

**User-facing:** any trigger with a macro in its Description shows the expanded value again on the default query. Behaviour with "Item value at problem time" enabled is unchanged, and the default path still sends no `history.get`.

### Why the parameter is omitted rather than sent as false

Zabbix expands the comment whenever the `expandComment` key is present, regardless of its value. Sending `false` on the opt-in path would let Zabbix expand the comment with the item's current value before the plugin sees it, leaving the per-problem expansion nothing to replace.

### DIFF
--- a/src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
+++ b/src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
@@ -632,12 +632,21 @@ export class ZabbixAPIConnector {
     return this.request('problem.get', params).then(utils.mustArray);
   }
 
-  async getTriggersByIds(triggerids: string[]) {
+  /**
+   * `expandComment` asks Zabbix to expand macros in the trigger comment, which the
+   * Problems panel renders as "Description". Pass false only when the caller expands
+   * those macros itself per problem, because that needs the raw template.
+   *
+   * The parameter must be omitted rather than sent as false. Zabbix treats the key
+   * as present-means-true and expands the comment even when the value is false.
+   */
+  async getTriggersByIds(triggerids: string[], expandComment = true) {
     const params = {
       output: 'extend',
       triggerids: triggerids,
       expandDescription: true,
       expandData: true,
+      ...(expandComment ? { expandComment: true } : {}),
       expandExpression: true,
       monitored: true,
       skipDependent: true,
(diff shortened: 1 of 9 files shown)


src/datasource/zabbix/connectors/zabbix_api/zabbixAPIConnector.ts
```diff
@@ -632,12 +632,21 @@ export class ZabbixAPIConnector {
     return this.request('problem.get', params).then(utils.mustArray);
   }
 
-  async getTriggersByIds(triggerids: string[]) {
+  /**
+   * `expandComment` asks Zabbix to expand macros in the trigger comment, which the
+   * Problems panel renders as "Description". Pass false only when the caller expands
```
fabrizio-grafana, line 637: > Pass false only when the caller expands those macros itself per problem, because that needs the raw template.

Hmm this sentence is not clear to me.

src/datasource/zabbix/zabbix.test.ts
```diff
@@ -229,6 +229,18 @@ describe('Zabbix', () => {
 
       expect(zabbix.zabbixAPI.getHistory).toHaveBeenCalled();
     });
+
+    it('asks Zabbix to expand trigger comments when the history lookup is off', async () => {
+      await zabbix.getProblemsHistory('group.*', 'host.*', 'app.*', undefined, {});
```
fabrizio-grafana, line 234: Do you think we can add unit tests for `getProblems` too?

fabrizio-grafana (approved):
A few nit comments, but overall looks good.

